PR #1, opened 2022-11-15
feat: added SDK config 

Shahroz16:
closes: https://github.com/customerio/issues/issues/8530

### DIFF
--- a/android/src/main/kotlin/io/customer/customer_io/CustomerIoPlugin.kt
+++ b/android/src/main/kotlin/io/customer/customer_io/CustomerIoPlugin.kt
@@ -81,11 +81,8 @@ class CustomerIOPlugin : FlutterPlugin, MethodCallHandler {
     private fun getUserAgentClient(packageConfig: Map<String, Any?>?): Client {
         val sourceSDKVersion = packageConfig?.getProperty<String>(
             Keys.PackageConfig.SOURCE_SDK_VERSION
-        )?.takeIfNotBlank() ?: packageConfig?.getProperty<String>(
-            Keys.PackageConfig.SOURCE_SDK_VERSION_COMPAT
         )?.takeIfNotBlank() ?: "n/a"
-        // TODO: change it to flutter
-        return Client.ReactNative(sdkVersion = sourceSDKVersion)
+        return Client.Flutter(sdkVersion = sourceSDKVersion)
     }
 
     private fun CustomerIO.Builder.setupConfig(config: Map<String, Any?>?): CustomerIO.Builder {
--- a/android/src/main/kotlin/io/customer/customer_io/CustomerIoPlugin.kt
+++ b/android/src/main/kotlin/io/customer/customer_io/CustomerIoPlugin.kt
@@ -57,9 +57,7 @@ class CustomerIoPlugin : FlutterPlugin, MethodCallHandler {
             val region = configData.getProperty<String>(
                 Keys.Environment.REGION
             )?.takeIfNotBlank().toRegion()
-
-            logger.info("Config: $configData")
-
+            
             CustomerIO.Builder(
                 siteId = siteId,
                 apiKey = apiKey,
--- a/android/src/main/kotlin/io/customer/customer_io/CustomerIoPlugin.kt
+++ b/android/src/main/kotlin/io/customer/customer_io/CustomerIoPlugin.kt
@@ -1,7 +1,14 @@
 package io.customer.customer_io
 
+import android.app.Application
+import android.content.Context
 import androidx.annotation.NonNull
-
+import io.customer.customer_io.constant.Keys
+import io.customer.customer_io.extension.*
+import io.customer.sdk.CustomerIO
+import io.customer.sdk.CustomerIOShared
+import io.customer.sdk.data.store.Client
+import io.customer.sdk.util.Logger
 import io.flutter.embedding.engine.plugins.FlutterPlugin
 import io.flutter.plugin.common.MethodCall
 import io.flutter.plugin.common.MethodChannel
@@ -9,27 +16,97 @@ import io.flutter.plugin.common.MethodChannel.MethodCallHandler
 import io.flutter.plugin.common.MethodChannel.Result
 
 /** CustomerIoPlugin */
-class CustomerIoPlugin: FlutterPlugin, MethodCallHandler {
-  /// The MethodChannel that will the communication between Flutter and native Android
-  ///
-  /// This local reference serves to register the plugin with the Flutter Engine and unregister it
-  /// when the Flutter Engine is detached from the Activity
-  private lateinit var channel : MethodChannel
+class CustomerIoPlugin : FlutterPlugin, MethodCallHandler {
+    /// The MethodChannel that will the communication between Flutter and native Android
+    ///
+    /// This local reference serves to register the plugin with the Flutter Engine and unregister it
+    /// when the Flutter Engine is detached from the Activity
+    private lateinit var channel: MethodChannel
+    private lateinit var context: Context
 
-  override fun onAttachedToEngine(@NonNull flutterPluginBinding: FlutterPlugin.FlutterPluginBinding) {
-    channel = MethodChannel(flutterPluginBinding.binaryMessenger, "customer_io")
-    channel.setMethodCallHandler(this)
-  }
+    private val logger: Logger
+        get() = CustomerIOShared.instance().diGraph.logger
 
-  override fun onMethodCall(@NonNull call: MethodCall, @NonNull result: Result) {
-    if (call.method == "getPlatformVersion") {
-      result.success("Android ${android.os.Build.VERSION.RELEASE}")
-    } else {
-      result.notImplemented()
+    override fun onAttachedToEngine(@NonNull flutterPluginBinding: FlutterPlugin.FlutterPluginBinding) {
+        context = flutterPluginBinding.applicationContext
+        channel = MethodChannel(flutterPluginBinding.binaryMessenger, "customer_io")
+        channel.setMethodCallHandler(this)
     }
-  }
 
-  override fun onDetachedFromEngine(@NonNull binding: FlutterPlugin.FlutterPluginBinding) {
-    channel.setMethodCallHandler(null)
-  }
+    override fun onMethodCall(@NonNull call: MethodCall, @NonNull result: Result) {
+        when (call.method) {
+            "getPlatformVersion" -> {
+                result.success("Android-${android.os.Build.VERSION.RELEASE}")
+                CustomerIO.instance().identify("flutt-man")
+            }
+            "initialize" -> {
+                initialize(call, result)
+            }
+            else -> {
+                result.notImplemented()
+            }
+        }
+    }
+
+
+    private fun initialize(call: MethodCall, result: Result) {
+        try {
+            val application: Application = context.applicationContext as Application
+            val configData = call.argument<Map<String, Any>>("config") ?: emptyMap()
+            val siteId = configData.getString(Keys.Environment.SITE_ID)
+            val apiKey = configData.getString(Keys.Environment.API_KEY)
+            val region = configData.getProperty<String>(
+                Keys.Environment.REGION
+            )?.takeIfNotBlank().toRegion()
+
+            CustomerIO.Builder(
+                siteId = siteId,
+                apiKey = apiKey,
+                region = region,
+                appContext = application,
+            ).apply {
+                setClient(client = getUserAgentClient(packageConfig = configData))
+                setupConfig(configData)
+            }.build()
+            logger.info("Customer.io instance initialized successfully")
+            result.success(true)
+        } catch (e: Exception) {
+            logger.error("Failed to initialize Customer.io instance from app, ${e.message}")
+            result.error("FlutterSegmentException", e.localizedMessage, null);
+        }
+    }
+
+    private fun getUserAgentClient(packageConfig: Map<String, Any?>?): Client {
+        val sourceSDKVersion = packageConfig?.getProperty<String>(
+            Keys.PackageConfig.SOURCE_SDK_VERSION
+        )?.takeIfNotBlank() ?: packageConfig?.getProperty<String>(
+            Keys.PackageConfig.SOURCE_SDK_VERSION_COMPAT
+        )?.takeIfNotBlank() ?: "n/a"
+        // TODO: change it to flutter
+        return Client.ReactNative(sdkVersion = sourceSDKVersion)
+    }
+
+    private fun CustomerIO.Builder.setupConfig(config: Map<String, Any?>?): CustomerIO.Builder {
+        if (config == null) return this
+
+        val logLevel = config.getProperty<String>(Keys.Config.LOG_LEVEL).toCIOLogLevel()
+        setLogLevel(level = logLevel)
+        config.getProperty<String>(Keys.Config.TRACKING_API_URL)?.takeIfNotBlank()?.let { value ->
+            setTrackingApiURL(value)
+        }
+        config.getProperty<Boolean>(Keys.Config.AUTO_TRACK_DEVICE_ATTRIBUTES)?.let { value ->
+            autoTrackDeviceAttributes(shouldTrackDeviceAttributes = value)
+        }
+        config.getProperty<Double>(Keys.Config.BACKGROUND_QUEUE_MIN_NUMBER_OF_TASKS)?.let { value ->
+            setBackgroundQueueMinNumberOfTasks(backgroundQueueMinNumberOfTasks = value.toInt())
+        }
+        config.getProperty<Double>(Keys.Config.BACKGROUND_QUEUE_SECONDS_DELAY)?.let { value ->
+            setBackgroundQueueSecondsDelay(backgroundQueueSecondsDelay = value)
+        }
+        return this
+    }
+
+    override fun onDetachedFromEngine(@NonNull binding: FlutterPlugin.FlutterPluginBinding) {
+        channel.setMethodCallHandler(null)
+    }
 }
--- a/android/src/main/kotlin/io/customer/customer_io/CustomerIoPlugin.kt
+++ b/android/src/main/kotlin/io/customer/customer_io/CustomerIoPlugin.kt
@@ -15,13 +15,16 @@ import io.flutter.plugin.common.MethodChannel
 import io.flutter.plugin.common.MethodChannel.MethodCallHandler
 import io.flutter.plugin.common.MethodChannel.Result
 
-/** CustomerIoPlugin */
-class CustomerIoPlugin : FlutterPlugin, MethodCallHandler {
+/**
+ * Android implementation of plugin that will let Flutter developers to
+ * interact with a Android platform
+ * */
+class CustomerIOPlugin : FlutterPlugin, MethodCallHandler {
     /// The MethodChannel that will the communication between Flutter and native Android
     ///
     /// This local reference serves to register the plugin with the Flutter Engine and unregister it
     /// when the Flutter Engine is detached from the Activity
-    private lateinit var channel: MethodChannel
+    private lateinit var flutterCommunicationChannel: MethodChannel
     private lateinit var context: Context
 
     private val logger: Logger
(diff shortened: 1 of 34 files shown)


ios/Classes/SwiftCustomerIoPlugin.swift
```diff
@@ -1,14 +1,65 @@
...
+        
+        CustomerIO.initialize(siteId: siteId, apiKey: apiKey, region: Region.from(regionStr: region))
+        setUserAgentClient(params: params)
+        setupConfig(params: params)
+    }
+    
+    private func setUserAgentClient(params : Dictionary<String, Any>){
+        let version = params[Keys.PackageConfig.version] as? String ?? "n/a"
+        let sdkSource = SdkWrapperConfig.Source.flutter
+        CustomerIO.config {
```
ami-aman, line 48: Set SDK configuration during initialisation only.

lib/customer_io_config.dart
```diff
@@ -0,0 +1,47 @@
+import 'customer_io_enums.dart';
+
+/// Configure plugin using CustomerIOConfig
+class CustomerIOConfig {
+  final String siteId;
+  final String apiKey;
+  Region region;
```
ami-aman, line 7: Region is also optional with a default value. I could be mistaken as this is dart and setting a default value could be done in a different way.

ios/Classes/Keys.swift
```diff
@@ -0,0 +1,32 @@
+//
+//  Keys.swift
+//  customer_io
+//
+//  Created by ShahrozAli on 11/11/22.
+//
```
ami-aman, line 6: Can be ignored right now, but we would need to remove the headers across all files before final release.

android/src/main/kotlin/io/customer/customer_io/CustomerIoPlugin.kt
```diff
@@ -1,35 +1,111 @@
...
+            logger.error("Failed to initialize Customer.io instance from app, ${e.message}")
+            result.error("FlutterSegmentException", e.localizedMessage, null);
+        }
+    }
+
+    private fun getUserAgentClient(packageConfig: Map<String, Any?>?): Client {
+        val sourceSDKVersion = packageConfig?.getProperty<String>(
+            Keys.PackageConfig.SOURCE_SDK_VERSION
+        )?.takeIfNotBlank() ?: packageConfig?.getProperty<String>(
+            Keys.PackageConfig.SOURCE_SDK_VERSION_COMPAT
```
levibostian, line 82: I am not sure what `Keys` is. 

Is this code saying that Flutter customers will set in their apps the value for `SOURCE_SDK_VERSION`? Or are we able to set that value ourselves?

android/src/main/kotlin/io/customer/customer_io/CustomerIoPlugin.kt
```diff
@@ -1,35 +1,111 @@
...
+        }
+        config.getProperty<Boolean>(Keys.Config.AUTO_TRACK_DEVICE_ATTRIBUTES)?.let { value ->
+            autoTrackDeviceAttributes(shouldTrackDeviceAttributes = value)
+        }
+        config.getProperty<Double>(Keys.Config.BACKGROUND_QUEUE_MIN_NUMBER_OF_TASKS)?.let { value ->
+            setBackgroundQueueMinNumberOfTasks(backgroundQueueMinNumberOfTasks = value.toInt())
+        }
+        config.getProperty<Double>(Keys.Config.BACKGROUND_QUEUE_SECONDS_DELAY)?.let { value ->
+            setBackgroundQueueSecondsDelay(backgroundQueueSecondsDelay = value)
+        }
```
levibostian, line 104: Is there a function that we can add to the native SDK or something to do this work for us? 

I feel this code is error-prone as we would need to remember to update this code each time that we update the SDK configuration object in the native SDKs. By housing the code in the native SDK, we have a higher chance of remembering.

android/src/main/kotlin/io/customer/customer_io/extension/MapExt.kt
```diff
@@ -0,0 +1,36 @@
+package io.customer.customer_io.extension
+
+import io.customer.sdk.CustomerIOShared
+
+@Throws(IllegalArgumentException::class)
+internal inline fun <reified T> Map<String, Any?>.getPropertyUnsafe(key: String): T {
```
levibostian, line 6:
```suggestion
internal inline fun <reified T> Map<String, Any?>.getValueUnsafe(key: String): T {
```

IMO, `value` reads easier since maps contain key value pairs.

android/src/main/kotlin/io/customer/customer_io/extension/TypeConversion.kt
```diff
@@ -0,0 +1,17 @@
...
+    else listOf(
+        Region.US,
+        Region.EU,
+    ).find { value -> value.code.equals(this, ignoreCase = true) } ?: fallback
+}
+
+internal fun String?.toCIOLogLevel(fallback: CioLogLevel = CioLogLevel.NONE): CioLogLevel {
+    return CioLogLevel.values().find { value -> value.name.equals(this, ignoreCase = true) }
+        ?: fallback
+}
```
levibostian, line 17: See PR comment in regards to housing code like this in the native SDKs instead of in wrapper SDK (in this case, Flutter).

example/lib/main.dart
```diff
@@ -1,10 +1,20 @@
...
-import 'package:flutter/services.dart';
 import 'package:customer_io/customer_io.dart';
+import 'package:customer_io/customer_io_config.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+
+void main() async {
+  WidgetsFlutterBinding.ensureInitialized();
+
+  await CustomerIo.initialize(
```
levibostian, line 11: Assuming that `await` in Dart is performing asynchronous operations, does `initialize()` need to be asynchronous since it's not in the native SDKs?

example/lib/main.dart
```diff
@@ -32,7 +41,7 @@ class _MyAppState extends State<MyApp> {
     // We also handle the message potentially returning null.
     try {
       platformVersion =
-          await _customerIoPlugin.getPlatformVersion() ?? 'Unknown platform version';
+          await CustomerIo.getPlatformVersion() ?? 'Unknown platform version';
```
levibostian, line 44:
```suggestion
          await CustomerIO.getPlatformVersion() ?? 'Unknown platform version';
```

example/lib/main.dart
levibostian:
```suggestion
  await CustomerIO.initialize(
```

example/lib/main.dart
```diff
@@ -1,10 +1,20 @@
...
+import 'package:flutter/services.dart';
+
+void main() async {
+  WidgetsFlutterBinding.ensureInitialized();
+
+  await CustomerIo.initialize(
+    config: CustomerIOConfig(
+      siteId: "",
+      apiKey: "",
+    ),
```
levibostian, line 15: Can we make the `CustomerIO.initalize()` function for the flutter SDK match the API of our native SDKs? 

By providing the siteId and apiKey as parameters of `initialize()` instead of inside of a config object? 

```suggestion
  CustomerIO.initialize(
      siteId: "",
      apiKey: "",
   ),
```

ios/Classes/CustomerIOExtensions.swift
```diff
@@ -0,0 +1,39 @@
...
+        case "error":
+            return .error
+        case "info":
+            return .info
+        case "debug":
+            return .debug
+        default:
+            return .error
+        }
+    }
```
levibostian, line 38: See PR comment in regards to housing code like this in the native SDKs instead of in wrapper SDK (in this case, Flutter).

ios/Classes/Keys.swift
levibostian: The Swift format and swift lint tools that we use on iOS SDK do this for us. 

We can configure git hooks like the Android SDK in this project that run the linter/formatter against this code?

ios/Classes/Keys.swift
```diff
@@ -0,0 +1,32 @@
...
+
+import Foundation
+
+struct Keys {
```
levibostian, line 10: This comment applies to both Android and iOS `Keys` objects. 

I see value in code like this in that we are containing strings into 1 place. However, I wonder if there is another idea for how to solve this problem? I think it sounds like a lot of work to maintain this file over time. 

We tried a similar idea in the RH iOS app and we found ourselves forgetting often to add constants in the file instead of having string values in the app. 

One idea is to move the code that uses these values into a util class that encapsulates the strings into 1 file:

```kotlin
// Before
// CustomerIOPlugin file
getUserAgent() {
  ...
  val sdkVersion = getValue(Keys.PackageConfig.version)

// After
// Code moved into SDKPackageUtil
class SDKPackageUtilImpl: SDKPackageUtil {
  val userAgent: Client {
    val sdkVersion = getValue("version")
  }
}
```

The code above isn't quite accurate but I hope you get the idea. Instead of having 1 file that contains a long list of random constants we can have the hard-coded strings encapsulated in 1 file that does the task that we need it to do. 

Seems more testable and maintainable to me in some ways.

android/src/main/kotlin/io/customer/customer_io/constant/Keys.kt
```diff
@@ -0,0 +1,24 @@
+package io.customer.customer_io.constant
+
+internal object Keys {
```
levibostian, line 3: See PR comment in regards to iOS `Keys` object

ios/Classes/SwiftCustomerIoPlugin.swift
```diff
@@ -1,14 +1,65 @@
...
-    let instance = SwiftCustomerIoPlugin()
-    registrar.addMethodCallDelegate(instance, channel: channel)
-  }
-
-  public func handle(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
-    result("iOS " + UIDevice.current.systemVersion)
-  }
+    
+    public static func register(with registrar: FlutterPluginRegistrar) {
+        let channel = FlutterMethodChannel(name: "customer_io", binaryMessenger: registrar.messenger())
```
levibostian, line 9: See PR comment on `"customer_io"`

ios/Classes/SwiftCustomerIoPlugin.swift
```diff
@@ -1,14 +1,65 @@
...
+        let instance = SwiftCustomerIoPlugin()
+        registrar.addMethodCallDelegate(instance, channel: channel)
+    }
+    
+    public func handle(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
+        switch(call.method) {
+        case "getPlatformVersion":
+            result("iOS " + UIDevice.current.systemVersion)
```
levibostian, line 17: Is this used for your testing?

ios/Classes/SwiftCustomerIoPlugin.swift
```diff
@@ -1,14 +1,65 @@
...
+            $0.autoTrackDeviceAttributes = params[Keys.Config.autoTrackDeviceAttributes] as! Bool
+            $0.logLevel = CioLogLevel.from(for: params[Keys.Config.logLevel] as! String)
+            $0.autoTrackPushEvents = params[Keys.Config.autoTrackPushEvents] as! Bool
+            $0.backgroundQueueMinNumberOfTasks = params[Keys.Config.backgroundQueueMinNumberOfTasks] as! Int
+            $0.backgroundQueueSecondsDelay = params[Keys.Config.backgroundQueueSecondsDelay] as! Seconds
+            if let trackingApiUrl = params[Keys.Config.trackingApiUrl] as? String, !trackingApiUrl.isEmpty {
+                $0.trackingApiUrl = trackingApiUrl
+            }
+        }
+    }
```
levibostian, line 64: See PR comment in regards to housing code like this in the native SDKs instead of in wrapper SDK (in this case, Flutter).

ios/customer_io.podspec
```diff
@@ -15,7 +15,8 @@ A plugin for Customer.io
   s.source           = { :path => '.' }
   s.source_files = 'Classes/**/*'
   s.dependency 'Flutter'
-  s.platform = :ios, '9.0'
+  s.dependency 'CustomerIO/Tracking'
```
levibostian, line 18: Is this file used for the example Flutter iOS app? Or is this the file used for us to push the flutter SDK to cocoapods so our flutter customers can use it in their app? 

If it's the cocoapods file that we use for publishing, I would like to make some changes to the file to match the naming style of our iOS SDK so the SDK is easier to find.

lib/customer_io.dart
```diff
@@ -1,8 +1,18 @@
-
+import 'customer_io_config.dart';
 import 'customer_io_platform_interface.dart';
 
 class CustomerIo {
-  Future<String?> getPlatformVersion() {
-    return CustomerIoPlatform.instance.getPlatformVersion();
+  const CustomerIo._();
```
levibostian, line 5: 🤔 What is this for? Curious....

lib/customer_io.dart
```diff
@@ -1,8 +1,18 @@
...
+
+  static CustomerIOPlatform get _customerIO => CustomerIOPlatform.instance;
+
+  static Future<String?> getPlatformVersion() {
+    return _customerIO.getPlatformVersion();
+  }
+
+  static Future<void> initialize({
```
levibostian, line 13: Ah, this must be why the other Dart file said `await CustomerIO.initialize()`. 

Does this function have to be a `Future` since the native SDK `initialize()` function isn't async?

lib/customer_io_config.dart
```diff
@@ -0,0 +1,47 @@
+import 'customer_io_enums.dart';
+
+/// Configure plugin using CustomerIOConfig
+class CustomerIOConfig {
+  final String siteId;
+  final String apiKey;
```
levibostian, line 6: Can we move these 2 properties out of the config object and into parameters of Dart's `CustomerIO.initialize()` function? To match the native SDKs?

lib/customer_io_config.dart
```diff
@@ -0,0 +1,47 @@
+import 'customer_io_enums.dart';
+
+/// Configure plugin using CustomerIOConfig
+class CustomerIOConfig {
```
levibostian, line 4: Ah, bummer. Looks like we might need to keep a Dart equivalent to the native SDK `CustomerIOConfig` maintained? 

Sounds like a lot of work that every time that we add a value to the Android or iOS SDK config object, we need to update it across RN SDK and Flutter SDK. But, I am not sure if there is a way around this?

ios/Classes/Keys.swift
Shahroz16: Yup will move on to `lint/hooks/ci/cd` after the basic implementation is done. Since i am the only one working, i dont get any conflicts as such either.

android/src/main/kotlin/io/customer/customer_io/CustomerIoPlugin.kt
```diff
@@ -1,35 +1,111 @@
...
+        context = flutterPluginBinding.applicationContext
+        channel = MethodChannel(flutterPluginBinding.binaryMessenger, "customer_io")
+        channel.setMethodCallHandler(this)
+    }
+
+    override fun onMethodCall(@NonNull call: MethodCall, @NonNull result: Result) {
+        when (call.method) {
+            "getPlatformVersion" -> {
+                result.success("Android-${android.os.Build.VERSION.RELEASE}")
+            }
```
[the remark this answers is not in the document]
Shahroz16: just for internal testing, will remove it afterwards.

ios/customer_io.podspec
Shahroz16: This file is used by iOS internally only, it fetches CustomerIO native SDK for us that we make native bindings with, but i'll dig down more once i start looking into publishing the flutter plugins.

lib/customer_io.dart
Shahroz16: This is how you make `singleton` classes in dart

lib/customer_io_config.dart
Shahroz16: Yup, that's the only way, we have to keep these wrapper updated with native SDK

android/src/main/kotlin/io/customer/customer_io/CustomerIoPlugin.kt
```diff
@@ -1,35 +1,111 @@
...
+    /// when the Flutter Engine is detached from the Activity
+    private lateinit var channel: MethodChannel
+    private lateinit var context: Context
+
+    private val logger: Logger
+        get() = CustomerIOShared.instance().diGraph.logger
+
+    override fun onAttachedToEngine(@NonNull flutterPluginBinding: FlutterPlugin.FlutterPluginBinding) {
+        context = flutterPluginBinding.applicationContext
+        channel = MethodChannel(flutterPluginBinding.binaryMessenger, "customer_io")
```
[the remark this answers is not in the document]
Shahroz16: Its the channel name, we can have multiple channels to communicate with the native platform and flutter, also the convention in dart is to use "_" rather than "."

lib/customer_io.dart
Shahroz16: This doesn't have to be `Future ` but it's a common practice in plugin development that I have noticed, they return back with success/failures. 

I believe it's for the sake of state management as if they want to perform all initializations first and then move on to the other screens, native platforms provide it themselves, but maybe Cross platform doesn't

android/src/main/kotlin/io/customer/customer_io/CustomerIoPlugin.kt
```diff
@@ -1,35 +1,114 @@
...
+            logger.error("Failed to initialize Customer.io instance from app, ${e.message}")
+            result.error("FlutterSegmentException", e.localizedMessage, null);
+        }
+    }
+
+    private fun getUserAgentClient(packageConfig: Map<String, Any?>?): Client {
+        val sourceSDKVersion = packageConfig?.getProperty<String>(
+            Keys.PackageConfig.SOURCE_SDK_VERSION
+        )?.takeIfNotBlank() ?: packageConfig?.getProperty<String>(
+            Keys.PackageConfig.SOURCE_SDK_VERSION_COMPAT
```
mrehan27, line 85: We probably don't need `SOURCE_SDK_VERSION_COMPAT` as it was added for supporting older attributes, which aren't needed for Flutter 👀

android/src/main/kotlin/io/customer/customer_io/CustomerIoPlugin.kt
mrehan27: I think we can support initialization through maps in Native SDKs to help in this for both React Native and Flutter. But we need to be careful and not rename any key used by these wrappers or change its type to avoid breaking this.

Shahroz16: 👍 Agreed, something that we should definitely focus later on.

android/src/main/kotlin/io/customer/customer_io/CustomerIoPlugin.kt
```diff
@@ -1,35 +1,114 @@
...
+            setBackgroundQueueMinNumberOfTasks(backgroundQueueMinNumberOfTasks = value.toInt())
+        }
+        config.getProperty<Double>(Keys.Config.BACKGROUND_QUEUE_SECONDS_DELAY)?.let { value ->
+            setBackgroundQueueSecondsDelay(backgroundQueueSecondsDelay = value)
+        }
+        return this
+    }
+
+    override fun onDetachedFromEngine(@NonNull binding: FlutterPlugin.FlutterPluginBinding) {
+        flutterCommunicationChannel.setMethodCallHandler(null)
```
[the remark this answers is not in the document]
Shahroz16: That's a good point, i will look more into it.

matt-frizzell (approved):
Looks good enough to push through on

android/src/main/kotlin/io/customer/customer_io/CustomerIoPlugin.kt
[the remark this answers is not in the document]
Shahroz16: Checked the documentation, and since the context is a private member its lifecycle is bound to the `plugin` lifecycle. 

And in this method, it will release all the resources held by this plugin.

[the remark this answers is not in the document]
mrehan27: Great, thanks for confirming and sharing the updates 👍🏻

mrehan27 (approved):
LGTM 👍🏻

PR #2, opened 2022-11-29
feat: tracking and in-app added

Shahroz16:


android/src/main/kotlin/io/customer/customer_io/CustomerIoPlugin.kt
```diff
@@ -19,7 +20,7 @@ import io.flutter.plugin.common.MethodChannel.Result
  * Android implementation of plugin that will let Flutter developers to
  * interact with a Android platform
  * */
-class CustomerIOPlugin : FlutterPlugin, MethodCallHandler {
+class CustomerIoPlugin : FlutterPlugin, MethodCallHandler {
```
Shahroz16, line 23: Had to revert back to `CustomerIoPlugin` because it's auto-generated and if you update it, it stops communicating to native bindings.

mrehan27 (approved):
LGTM 👍🏻

PR #17, opened 2023-01-10
feat: added missing methods

Shahroz16:


ios/Classes/SwiftCustomerIoPlugin.swift
```diff
@@ -119,6 +127,28 @@ public class SwiftCustomerIoPlugin: NSObject, FlutterPlugin {
         CustomerIO.shared.profileAttributes = attributes
     }
     
+    private func registerDeviceToken(params : Dictionary<String, AnyHashable>){
+        guard let token = params[Keys.Tracking.token] as? String
+        else {
+            return
+        }
+        
+        CustomerIO.shared.registerDeviceToken(token)
```
mrehan27, line 136: I ain't an iOS expert 😅 But doesn't if statement looks better here? e.g.
```
if let token = params[Keys.Tracking.token] as? String {
    CustomerIO.shared.registerDeviceToken(token)
}
```
Mainly because I think the method is too small for early returns.

ios/Classes/SwiftCustomerIoPlugin.swift
```diff
@@ -119,6 +127,28 @@ public class SwiftCustomerIoPlugin: NSObject, FlutterPlugin {
...
+              let deviceToken = params[Keys.Tracking.deliveryToken] as? String,
+              let metricEvent = params[Keys.Tracking.metricEvent] as? String,
+              let event = Metric.getEvent(from: metricEvent)
+        else {
+            return
+        }
+        
+        CustomerIO.shared.trackMetric(deliveryID: deliveryId,
+                                      event: event,
+                                      deviceToken: deviceToken)
```
mrehan27, line 150: Same as above comment

mrehan27 (approved):
Just a few small queries, rest looks good 👍🏻

ios/Classes/SwiftCustomerIoPlugin.swift
Shahroz16: haha! I don't have any preference but I saw this being followed in RN iOS, so kept the same format. 😄

android/src/main/kotlin/io/customer/customer_io/CustomerIoPlugin.kt
```diff
@@ -142,9 +154,28 @@ class CustomerIoPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
...
+        CustomerIO.instance().registerDeviceToken(token)
+    }
+
+    private fun trackMetric(params: Map<String, Any>) {
+        val deliveryId = params.getString(Keys.Tracking.DELIVERY_ID)
+        val deliveryToken = params.getString(Keys.Tracking.DELIVERY_TOKEN)
+        val event = MetricEvent.getEvent(params.getProperty<String>(Keys.Tracking.METRIC_EVENT))
+
+        if (event == null) {
+            logger.info("metric event type isn't correct")
```
levibostian, line 168: Can we move this logic of validating the metric event type into the dart code? 

I understand that there is a chance that `MetricEvent.getEvent()` can still return nil in this Kotlin code layer, but if it does, I think it would be considered an SDK bug instead of an error for the customer. Therefore, I would expect to see the log statement be `logger.error("metric event null. Possible issue with SDK code? Given: ${params.getProperty<String>(Keys.Tracking.METRIC_EVENT)}")`

android/src/main/kotlin/io/customer/customer_io/CustomerIoPlugin.kt
```diff
@@ -142,9 +154,28 @@ class CustomerIoPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
...
+
+        CustomerIO.instance().trackMetric(
+            deliveryID = deliveryId, deviceToken = deliveryToken, event = event
+        )
+    }
+
     private fun setDeviceAttributes(params: Map<String, Any>) {
-        val attributes =
-            params.getProperty<Map<String, Any>>(Keys.Tracking.ATTRIBUTES) ?: emptyMap()
+        val attributes = params.getProperty<Map<String, Any>>(Keys.Tracking.ATTRIBUTES) ?: return
```
levibostian, line 178: Since it's relevant to this PR, I wanted to put a comment here. 

This is an example of using the guard design pattern in Kotlin 😄

lib/customer_io_method_channel.dart
```diff
@@ -117,6 +118,26 @@ class CustomerIOMethodChannel extends CustomerIOPlatform {
...
+      final payload = {
+        TrackingConsts.deliveryId: deliveryID,
+        TrackingConsts.deliveryToken: deviceToken,
+        TrackingConsts.metricEvent: event.name,
+      };
+      methodChannel.invokeMethod(MethodConsts.trackMetric, payload);
+    } on PlatformException catch (exception) {
+      if (kDebugMode) {
+        print(exception);
+      }
```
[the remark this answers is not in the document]
Shahroz16: That's a good point, I'll create a method in the same class.

test/customer_io_test.dart
```diff
@@ -56,6 +57,16 @@ class MockCustomerIoPlatform
...
+  @override
+  void registerDeviceToken({required String deviceToken}) {
+    // TODO: implement registerDeviceToken
+  }
+
+  @override
+  void trackMetric({required String deliveryID, required String deviceToken, required MetricEvent event}) {
+    // TODO: implement trackMetric
+  }
 }
```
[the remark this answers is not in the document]
Shahroz16: Added comment for it.

android/src/main/kotlin/io/customer/customer_io/CustomerIoPlugin.kt
Shahroz16: We do perform the validation in the `dart` file and are certain that the `event` sent will be catered by the SDK. 

I can update the log statement.

android/src/main/kotlin/io/customer/customer_io/CustomerIoPlugin.kt
```diff
@@ -142,9 +154,28 @@ class CustomerIoPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
         }
     }
 
+    private fun registerDeviceToken(params: Map<String, Any>) {
```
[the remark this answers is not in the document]
Shahroz16: Mostly for consistency, Since other methods are sending `Map` we have now a utility method that verifies the map and its keys. 

Also, if we ever wanted to send more params along with `token` we won't have to change the signature of the method or implementation but rather just add another `pair` to the map. 

But I agree that this looks like overkill and I might deviate away from this as well if we get 1 more method that just takes 1 `string` parameter.

levibostian (approved):
Looks good besides 1 new PR conversation I started in `lib/customer_io_method_channel.dart` file.

lib/customer_io_method_channel.dart
```diff
@@ -91,11 +89,10 @@ class CustomerIOMethodChannel extends CustomerIOPlatform {
         TrackingConsts.identifier: identifier,
         TrackingConsts.attributes: attributes
       };
+      methodChannel.invokeMethod("method", "");
```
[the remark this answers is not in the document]
Shahroz16: Oh yes, such a good carch, that was added by mistake while I tried to convert `Map` to `String`. This line is leftover from there, will remove it. And create another PR with changes.

PR #18, opened 2023-01-10
build: setup ios example to use locally installed iOS SDK if found

levibostian:
Setup iOS sample app to install iOS SDK locally if found. 

This is a similar setup that we use in Remote Habits iOS: Automatically install locally installed iOS SDK *if the iOS SDK directory exists at a certain path*. 

So, if you clone your git repos in the same directory on your computer:
```
customerio-flutter/
customerio-ios/
```

Then Flutter SDK will use the locally installed iOS SDK.


Shahroz16 (changes requested):
Added comments on the slack thread.

Shahroz16:
Closing this as this has been tackled.

PR #19, opened 2023-01-21
fix: in-app remove gist org id

Shahroz16:


android/src/main/kotlin/io/customer/customer_io/CustomerIoPlugin.kt
```diff
@@ -153,6 +151,9 @@ class CustomerIoPlugin : FlutterPlugin, MethodCallHandler {
         val organizationId = configData.getProperty<String>(
             Keys.Environment.ORGANIZATION_ID
         )?.takeIfNotBlank()
```
mrehan27, line 153: I would suggest to completely remove `organizationId` from native files and rely on `enableInApp` only, so it is convenient to generate warnings for desired platform and do the migrations for breaking changes later. I followed the [same idea in react native](https://github.com/customerio/customerio-reactnative/blob/8b97125d471f6a332722cee7963f1e48de29ef5b/src/CustomerioTracking.tsx#L54-L58) as well.

.gitignore
```diff
@@ -596,4 +596,5 @@ healthchecksdb
 # Backup folder for Package Reference Convert tool in Visual Studio 2017
 MigrationBackup/
 
-# End of https://www.gitignore.io/api/git,dart,flutter,intellij,webstorm,visualstudio
\ No newline at end of file
+# End of https://www.gitignore.io/api/git,dart,flutter,intellij,webstorm,visualstudio
+*.env
```
mrehan27, line 600: Is this deliberate? 👀

mrehan27 (changes requested):

.gitignore
Shahroz16: the order is wrong :D the addition is deliberate, it should be before the "End" comment.

mrehan27 (approved):

PR #21, opened 2023-02-08
feat: added in-app events listener

Shahroz16:


mrehan27 (commented):
Just minor queries, rest looks good to me

mrehan27 (approved):

android/build.gradle
```diff
@@ -44,6 +44,15 @@ android {
...
     }
+
+    tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
+        kotlinOptions {
+            freeCompilerArgs += [
+                    '-Xopt-in=kotlin.RequiresOptIn',
+                    '-Xopt-in=io.customer.base.internal.InternalCustomerIOApi',
+            ]
+        }
+    }
```
levibostian, line 55: Could you leave a comment saying why this is added? Help future code readers understand the decision behind adding this.

android/src/main/kotlin/io/customer/customer_io/CustomerIoPlugin.kt
```diff
@@ -22,17 +31,34 @@ import io.flutter.plugin.common.MethodChannel.Result
...
  * */
-class CustomerIoPlugin : FlutterPlugin, MethodCallHandler {
+class CustomerIoPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
     /// The MethodChannel that will the communication between Flutter and native Android
     ///
     /// This local reference serves to register the plugin with the Flutter Engine and unregister it
     /// when the Flutter Engine is detached from the Activity
     private lateinit var flutterCommunicationChannel: MethodChannel
     private lateinit var context: Context
+    private var activity: Activity? = null
```
levibostian, line 41: I think this might cause a memory leak to create a strong reference to an Activity? 

I know that the property is being set and unset based off of `ActivityAware` listener events, but can we rely on that? Or, do we need to make this property a weak reference? 

Is there a solution where we don't need to involve an Activity at all? Avoid this potentially being a problem? 

Assuming there is not a good solution not using an Activity, I am OK with the code proposed if you feel confident in it.

android/src/main/kotlin/io/customer/customer_io/CustomerIoPlugin.kt
```diff
@@ -147,23 +173,29 @@ class CustomerIoPlugin : FlutterPlugin, MethodCallHandler {
...
-            setClient(client = getUserAgentClient(packageConfig = configData))
-            setupConfig(configData)
             addCustomerIOModule(module = configureModuleMessagingPushFCM(configData))
             if (enableInApp == true) {
                 addCustomerIOModule(
-                    module = ModuleMessagingInApp()
+                    module = ModuleMessagingInApp(
+                        config = MessagingInAppModuleConfig.Builder()
+                            .setEventListener(CustomerIOInAppEventListener { method, args ->
+                                this@CustomerIoPlugin.activity?.runOnUiThread {
```
levibostian, line 194: This is the main reason why I bring up the strong reference to the Activity and a memory leak. The activity is used inside of the event listener object. 

So, if the CIO SDK keeps a strong reference to the in-app event listener --> the event listener keeps a strong reference to the Activity -> the Activity might not get freed from memory in this class? 

I am unsure if this is true?

example/credentials.env
```diff
@@ -0,0 +1,2 @@
+siteId=SITE_ID
```
levibostian, line 1: I think this file should:
1. be added to `.gitignore`
2. have a `credentials.env.example` file created that shows the expected format of the file. 

It would be pretty easy for someone to accidentally add this file to version control by accident with real creds.

lib/customer_io.dart
```diff
@@ -70,4 +73,15 @@ class CustomerIO {
...
     return _customerIO.setProfileAttributes(attributes: attributes);
   }
+
+  /// Subscribes to an in-app event listener.
+  ///
+  /// [onEvent] - A callback function that will be called every time an in-app event occurs.
+  /// The callback returns [InAppEvent].
+  ///
+  /// Returns a [StreamSubscription] that can be used to subscribe/unsubscribe from the event listener.
+  static StreamSubscription subscribeToInAppEventListener(
```
levibostian, line 83: Is this function an async method? I am unsure from the function signature but I see `dark:async` module imported into the file making me think it is. 

If so, if there a reason for making this function async? I think the code runs synchronously?

PR #24, opened 2023-02-16
fix: updated android dependency to auto update

Shahroz16:
Following iOS to all Android to auto-update minor updates for SDK. 

closes: https://github.com/customerio/issues/issues/9408#issuecomment-1431453510

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -58,7 +58,7 @@ android {
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
     // Customer.io SDK
-    def cioVersion = "3.3.0-beta.1"
+    def cioVersion = "[3.3,4.0)"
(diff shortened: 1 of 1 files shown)


levibostian (approved):

levibostian (changes requested):
The current Flutter `alpha` requires a beta deployment of the Android SDK. However, `[..)` syntax doesn't install pre-production deployments. So, this PR change might mean that the Android SDK `3.2.0` gets deployed. 

Merging this PR _might_ break the Flutter SDK that is deployed. So, I am going to reject for now. If you find that I am incorrect on this, go ahead and merge this. I approve if 3.3.0-beta.1 is being used by Flutter with this change.

Shahroz16:
@levibostian that's a very good call out, I am not going to wait on this, and verify it.

mrehan27 (approved):
The range is working fine in React Native SDK for beta releases as well. So looks good to me 👍🏻

PR #38, opened 2023-04-12
feat: process push notifications received outside CIO SDK

mrehan27:
Closes: https://github.com/customerio/issues/issues/9819

### Changes

- Added `customer_io_messaging_push` channel for push messaging method requests and platform implementation for push messaging module
- Added `CustomerIOPluginModule` for introducing module concept in Flutter SDK
- Added support to generate notification locally using `RemoteMessage` for FCM
- Extensions functions on `Map` for type casting, on `MethodCall` for returning value provided by native method and new constants required for the changes

### Sample Usages

For anyone receiving notifications outside Customer.io SDK, they can generate notifications locally using the following code snippet:

```
// For messages received when app is in foreground
CustomerIO.messagingPush().onMessageReceived(payload).then((handled) {
  // handled is true if notification was handled by Customer.io SDK; false otherwise
  return handled;
});

// For messages received when app is in background
CustomerIO.messagingPush().onBackgroundMessageReceived(payload).then((handled) {
  // handled is true if notification was handled by Customer.io SDK; false otherwise
  return handled;
});
```

where `payload` is map which contains notification payload received using FCM.

An example using FlutterFire would look like:

_For listening and processing messages received when app is in background_

```
Future<void> _firebaseMessagingBackgroundHandler(RemoteMessage message) async {
  await Firebase.initializeApp();
  CustomerIO.messagingPush().onBackgroundMessageReceived(message.toMap()).then((handled) {
    // handled is true if notification was handled by Customer.io SDK; false otherwise
    return handled;
  });
}

void main() async {
  // Initialize required SDKs
  FirebaseMessaging.onBackgroundMessage(_firebaseMessagingBackgroundHandler);
  // Run the app
}
```

_For listening and processing messages received when app is in foreground_

```
FirebaseMessaging.onMessage.listen((RemoteMessage message) {
  CustomerIO.messagingPush().onMessageReceived(message.toMap()).then((handled) {
    // handled is true if notification was handled by Customer.io SDK; false otherwise
    return handled;
  });
});
```


levibostian (approved):
I like the API designed (thanks for that in the PR description) and the implementation seems fine to me. 

The only part of the PR I hesitate on is the idea of `CustomerIOPluginModule` and this idea of having Flutter SDK modules map to a native Android module. 

I like the idea of keeping Flutter SDK push code encapsulated in it's own files and classes as you have done. But I am not understanding the value that we get from trying to map Flutter SDK to Android modules?

mrehan27:
Thanks @levibostian. I just wanted to make it easier for us to decide the correct file/class for adding new features. And mapping Flutter modules to native modules would probably be the easiest criteria. Also, if at any point, we make our wrapper SDKs modular like native and make adding in-app and push module optional, having them mapped to separate files will make the transition easier for us I think.

I'm still okay for defining any other criteria for splitting functionality into multiple files to keep the code readable and easier to maintain, but mapping them to native modules is probably the one I would still vote for.

lib/messaging_push/method_channel.dart
```diff
@@ -0,0 +1,48 @@
...
+  @override
+  Future<bool> onMessageReceived(Map<String, dynamic> message,
+      {bool handleNotificationTrigger = true}) {
+    if (Platform.isIOS) {
+      /// Since push notifications on iOS work fine with multiple notification
+      /// SDKs, we don't need to process them on iOS for now.
+      /// Resolving future to true makes it easier for callers to avoid adding
+      /// unnecessary platform specific checks.
+      return Future.value(true);
+    } else {
```
Shahroz16, line 26: not sure why we have `else` here if the `if` is returning in the first block?

Shahroz16 (approved):
Looks good, but need through testing, would appreciate if we can add test cases for the method channel as well

levibostian:
@mrehan27 in response to [your message here](https://github.com/customerio/customerio-flutter/pull/38#issuecomment-1506760436) I would prefer that we remove the plugin/module functionality from this PR. I suggest this for 2 reasons. 

1. I appreciate you thinking ahead and considering the possibility of breaking apart the Flutter SDK into modules like we do with our native SDKs. However, I prefer not to maintain this module code until we decide to implement that feature. 

2. Also, I don't want to make decisions today for this possible future feature that could limit our flexibility. By us deciding in this PR today how Flutter modules might be split up, that limits our decisions in the future for this feature. Maybe in the future we decide not to make push a separate module? Maybe we do? I see cons with making that decision today for our future selves. 

With all that being said, I appreciate you breaking it apart into files/classes. I think the API looks great as well.

mrehan27:
Thanks for sharing the concerns @levibostian. This still remains my secondary concern 😅 My main goal was to break it apart into files/classes based on some criteria so it is easier to take decisions for correct files in future as well. We don't have any other criteria for breaking down the functionality right now. And if we remove plugin/module functionality, this may cause most of the work in this PR to be re-done, which isn't ideal.

Since we all agree on the api design, we can remove plugin concept in future PRs if needed without changing public apis. For now, I would prefer sticking with the solution and not introduce any major changes in this PR.

FDuhen:
Any updates about this promising PR ? 
We're having issues using CIO alongside with Flutter Local Notifications, and from what I've understood, this update would help us a lot

mrehan27:
Hey @FDuhen. Thank you for reaching out. This is currently being tested and will be released once testing is complete.

PR #43, opened 2023-04-26
fix: missing opened metric on android 12 and above

Shahroz16:


android/build.gradle
```diff
@@ -58,7 +58,7 @@ android {
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
     // Customer.io SDK
-    def cioVersion = "[3.3,4.0)"
+    def cioVersion = "[3.4,4.0)"
```
mrehan27, line 61: Since this is range, we can probably update the lowest version to current latest? 🤔 

```suggestion
    def cioVersion = "[3.4.2,4.0)"
```

android/src/main/kotlin/io/customer/customer_io/CustomerIoPlugin.kt
```diff
@@ -214,7 +223,9 @@ class CustomerIoPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
             Keys.Environment.ENABLE_IN_APP
         )
 
-        CustomerIO.Builder(
+        val isLifecycleCallbacksRegistered = kotlin.runCatching { CustomerIO.instance() }.isSuccess
```
mrehan27, line 226: A small comment to this variable would be helpful

mrehan27 (approved):
Just minor suggestions, rest looks good

PR #51, opened 2023-05-30
feat: in-app dismiss support

Shahroz16:
closes: https://github.com/customerio/issues/issues/9932

Usage:
```
CustomerIO.messagingInApp().dismissMessage();
```

### DIFF
--- a/android/src/main/kotlin/io/customer/customer_io/messaginginapp/CustomerIOInAppMessaging.kt
+++ b/android/src/main/kotlin/io/customer/customer_io/messaginginapp/CustomerIOInAppMessaging.kt
@@ -1,0 +1,45 @@
+package io.customer.customer_io.messaginginapp
+
+import io.customer.customer_io.CustomerIOPluginModule
+import io.customer.customer_io.constant.Keys
+import io.customer.customer_io.invokeNative
+import io.customer.messaginginapp.di.inAppMessaging
+import io.customer.sdk.CustomerIO
+import io.flutter.embedding.engine.plugins.FlutterPlugin
+import io.flutter.plugin.common.MethodCall
+import io.flutter.plugin.common.MethodChannel
+
+/**
+ * Flutter module implementation for messaging in-app module in native SDKs. All functionality
+ * linked with the module should be placed here.
+ */
+internal class CustomerIOInAppMessaging(
+    pluginBinding: FlutterPlugin.FlutterPluginBinding,
+) : CustomerIOPluginModule, MethodChannel.MethodCallHandler {
+    override val moduleName: String = "InAppMessaging"
+    private val flutterCommunicationChannel: MethodChannel =
+        MethodChannel(pluginBinding.binaryMessenger, "customer_io_messaging_in_app")
+
+    override fun onAttachedToEngine() {
+        flutterCommunicationChannel.setMethodCallHandler(this)
+    }
+
+    override fun onDetachedFromEngine() {
+        flutterCommunicationChannel.setMethodCallHandler(null)
+    }
+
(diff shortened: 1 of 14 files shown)


lib/customer_io.dart
```diff
@@ -1,5 +1,7 @@
 import 'dart:async';
 
+import 'package:customer_io/messaging_in_app/platform_interface.dart';
+
```
mrehan27, line 4: I think we should keep imports consistent. e.g.
```
import 'messaging_in_app/platform_interface.dart';
import 'messaging_push/platform_interface.dart';
```
OR
```
import 'package:customer_io/messaging_in_app/platform_interface.dart';
import 'package:customer_io/messaging_push/platform_interface.dart';
```

I would prefer the first one.

ios/Classes/SwiftCustomerIoPlugin.swift
```diff
@@ -7,59 +7,63 @@ import CioMessagingInApp
...
+        instance.inAppMessagingChannelHandler = CusomterIOInAppMessaging(with: registrar)
     }
     
     deinit {
         self.methodChannel.setMethodCallHandler(nil)
+        self.inAppMessagingChannelHandler.detachFromEngine()
     }
     
     public func handle(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
         switch(call.method) {
```
mrehan27, line 26: I don't have anything in mind to suggest on this right now but this looks very tough to read on Github at least 😅 May be some formatting might help cleaning this up a little? 🤓

mrehan27 (approved):
Looks good overall. I would suggest we should update gradle range as well [like we did in react native](https://github.com/customerio/customerio-reactnative/pull/138/files#diff-3ab34fb625c346ec8377e7204444075ec41563ca1227c9a231df8c966386de89R5).

ios/Classes/SwiftCustomerIoPlugin.swift
mrehan27: Yes, that looks better 😅 Although I think it would be even better if it was formatted like follows:

```swift
    call.toNativeMethodCall(result: result) {
        initialize(params: $0)
    }
```

But if the current formatting is how swift formatting works then I'm fine with it and we can simply resolve this conversation 😄

android/src/main/kotlin/io/customer/customer_io/messaginginapp/CustomerIOInAppMessaging.kt
```diff
@@ -0,0 +1,45 @@
...
+    private val flutterCommunicationChannel: MethodChannel =
+        MethodChannel(pluginBinding.binaryMessenger, "customer_io_messaging_in_app")
+
+    override fun onAttachedToEngine() {
+        flutterCommunicationChannel.setMethodCallHandler(this)
+    }
+
+    override fun onDetachedFromEngine() {
+        flutterCommunicationChannel.setMethodCallHandler(null)
+    }
```
levibostian, line 29: Some of the code in this class are duplicate of what's in `CustomerIOPushMessaging`. 

Could we create an abstract class or something similar to avoid duplication? 

I want to avoid a potential bug in the future where we need to update the `onAttachedToEngine` implementation, for example, and forget to update the implementation in all classes in the SDK.

levibostian (approved):

android/src/main/kotlin/io/customer/customer_io/messaginginapp/CustomerIOInAppMessaging.kt
Shahroz16: We can cater to this in the later refactor, don't want to make it a part of this release as the refactor would effect `CustomerIOPushMessaging` as well.

PR #53, opened 2023-06-02
fix: hardcode iOS version to latest to avoid pods conflict

Shahroz16:


mrehan27:
@Shahroz16 I think we should move this to draft for now?

Shahroz16:
@mrehan27 yes we can.

levibostian:
Closing due to age. Please, re-open if you believe it should be.

PR #56, opened 2023-06-08
refactor: code optimizations

Shahroz16:


example/pubspec.yaml
```diff
@@ -17,7 +17,6 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
-  flutter_dotenv: ^5.0.2
```
Shahroz16, line 20: removing it because we will be adding a separate sample app in the same repo to demonstrate this

android/gradle/wrapper/gradle-wrapper.properties
mrehan27, line 1: Why there was no `gradle-wrapper.properties` file before? Is it not needed? 🤔

mrehan27 (approved):
LGTM 👍🏻

ami-aman (approved):

example/pubspec.yaml
levibostian: Thanks for the comment. I was going to ask why this change was made.

levibostian (approved):
Great change!

PR #65, opened 2023-07-21
fix: in-app messages not displaying for release builds on Android

levibostian:
Ships the latest Android SDK bug fix to RN customers.

# Tasks left to complete before merging this PR. 

- [x] Enable proguard in sample apps. After enabling proguard, push commit to this PR to assert proguard is enabled in the future for sample app builds. 
- [x] QA test sample app builds (that are now using proguard). Send device an in-app message. In-app message should appear on the device.


Shahroz16 (approved):
We might need to update version files manually because of flaky `sed` command

PR #60, opened 2023-06-29
ci: test publishing before merging PR to help prevent deploy failure

levibostian:
This is an attempt at preventing [deploy failures](https://github.com/customerio/customerio-flutter/actions/runs/5187051663/jobs/9348935326) by running similar commands deployment does before we merge PRs. 

This PR also added a setup-flutter action to make code more re-usable across workflow files.

PR #66, opened 2023-07-25
ci: updated ci pipeline and release workflow

Shahroz16:
- Change version update strategy 
- Added dry test before release from #60 
- Modified podspec dependency


.github/actions/setup-flutter/action.yml
```diff
@@ -0,0 +1,18 @@
+name: Setup 
+description: Setup runner to be able to build an app with Flutter. Not specific for Android or iOS.
+
+runs:
+  using: "composite"
+  steps:
+  # Install latest stable version of Flutter. Mostly because k-paxian/dart-package-publisher always uses the latest stable version of Flutter and we want the rest of the CI to match that.
```
mrehan27, line 7: Do you think it can break something in future if there is major release for Flutter and we haven't tested in with our SDK?

mrehan27 (approved):

.github/actions/setup-flutter/action.yml
Shahroz16: That is a good point, but it also make sure to keep us updated with the latest version and if our SDK doesn't break with it. We can revisit it and hardcode the version here if needed.

levibostian: I thought of this too, but I also noticed that `k-paxian/dart-package-publisher` doesn't let you define a flutter version. We used to run tests/lint with flutter version X but deploy (using `k-paxian/dart-package-publisher`) with flutter version Y and it caused deployments to fail. 

Therefore, no matter what we do, let's make sure to always use the same version of Flutter for the whole CI pipeline. At this time, because we can't specify flutter version during deployment, is the latest stable version.

levibostian (commented):
👍 great PR! Super excited to have this merged!

.github/workflows/test.yml
```diff
@@ -1,27 +1,30 @@
...
 
-      - name: Run tests
-        run: flutter test
\ No newline at end of file
+    - name: dry-run publishing to feel confident deployment will work after merging PR
+      uses: k-paxian/dart-package-publisher@v1.6
+      with:
+        credentialJson: ${{ secrets.CREDENTIAL_JSON }}
+        dryRunOnly: true 
+        flutter: true
```
levibostian, line 30: This code the way that it is will do nothing, unfortunately. 

[If you view the output in the CI](https://github.com/customerio/customerio-flutter/actions/runs/5656432153/job/15323447896#step:5:187), you will notice that dry-run exits early because the version number has not been changed yet. 

This is the reasoning why I was experimenting with [this hack](https://github.com/customerio/customerio-flutter/pull/60/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R25-R34) of incrementing the version of the package and modifying the changelog to trick flutter into thinking we were introducing a new version. Which, then will continue running the dry-run deployment instead of exiting early. 

I hope there is a better solution then what I came up with. However, it's not that complicated of a solution so if we have to use it, it's still an option.

levibostian (commented):
Oooopppps. I forgot to mention the dry-run deployment introduced by this PR has no effect. See my new comment I made.

.github/workflows/test.yml
Shahroz16: Good point, maybe we can do this before attempting to release. Let me think more about how this can be done.

PR #77, opened 2023-09-14
chore: sample app added

Shahroz16:


apps/amiapp_flutter/LICENSE
mrehan27, line 1: This isn't needed here now I think?

apps/amiapp_flutter/analysis_options.yaml
mrehan27, line 1: Is there an option to set only one linter for the repo (both sdk and sample app)?

mrehan27 (approved):

Issue #84, opened 2023-09-22
Refactor CustomerIO class to Enable Dependency Injection, Better Testing, and Flexibility


AristideVB:
### Description:

The current design of the `CustomerIO` class relies on static methods, making it difficult to write unit tests and achieve Dependency Injection. Converting static methods to instance methods will make the class more testable and flexible.

### Proposed Changes:

1. Convert static methods in `CustomerIO` to instance methods.
2. Continue to use the existing private constructor `const CustomerIO._();` to ensure that `CustomerIO` remains a singleton.

This approach will maintain the singleton behavior while offering greater flexibility for testing and modularization.

#### Before:

```dart
// Using static method
CustomerIO.subscribeToInAppEventListener(_handleInAppEvent);
```

#### After:

```dart
// Using instance method through singleton
final customerIO = CustomerIO.instance;
customerIO.subscribeToInAppEventListener(_handleInAppEvent);
```

### Practical Example with InAppMessagesBloc:

Suppose you have an `InAppMessagesBloc` that handles incoming in-app messages. In the current architecture, mocking `CustomerIO` for testing the `Bloc` is difficult.

#### Before:

```dart
class InAppMessagesBloc extends Bloc<InAppMessagesEvent, InAppMessagesState> {
  InAppMessagesBloc() : super(InAppMessagesInitial()) {
    on<InAppMessagesEventReceived>(_onInAppMessagesEventReceived);

    _inAppMessageStreamSubscription =
        CustomerIO.subscribeToInAppEventListener(_handleInAppEvent);
  }

  late StreamSubscription<dynamic> _inAppMessageStreamSubscription;
 // ...
}
```

#### After:

With the proposed changes, we can inject a mock or stubbed `CustomerIO` instance into the `InAppMessagesBloc` for better unit testing.

```dart
class InAppMessagesBloc extends Bloc<InAppMessagesEvent, InAppMessagesState> {
  InAppMessagesBloc(CustomerIO customerIO) : super(InAppMessagesInitial()) {
    on<InAppMessagesEventReceived>(_onInAppMessagesEventReceived);

    _inAppMessageStreamSubscription =
        customerIO.subscribeToInAppEventListener(_handleInAppEvent);
  }

  late StreamSubscription<dynamic> _inAppMessageStreamSubscription;
 // ...
}
```

### Benefits:

- Enhances testability by allowing for mocking or stubbing of methods.
- Allows for Dependency Injection in more modular codebases.
[...]

PR #85, opened 2023-09-22
feat: enable dependency injection for CustomerIO

AristideVB:
### Description:

This PR refactors the `CustomerIO` class to improve its testability, flexibility, and ease of use with Dependency Injection. The key changes made are converting the static methods to instance methods and introducing a singleton instance of `CustomerIO` for ease of use.

### Changes:

1. Convert static methods in `CustomerIO` to instance methods.
2. Introduce a singleton instance to maintain ease of use in existing implementations.

#### Before:

```dart
// Using static method
CustomerIO.subscribeToInAppEventListener(_handleInAppEvent);
```

#### After:

```dart
// Using instance method
final customerIO = CustomerIO.instance;
customerIO.subscribeToInAppEventListener(_handleInAppEvent);
```

#### Practical Usage in a Bloc:

```dart
class InAppMessagesBloc extends Bloc<InAppMessagesEvent, InAppMessagesState> {
  InAppMessagesBloc(CustomerIO customerIO) : super(InAppMessagesInitial()) {
    on<InAppMessagesEventReceived>(_onInAppMessagesEventReceived);

    _inAppMessageStreamSubscription =
        customerIO.subscribeToInAppEventListener(_handleInAppEvent);
  }

  late StreamSubscription<dynamic> _inAppMessageStreamSubscription;
}
```

### Benefits:

- Enables Dependency Injection, making the class more modular and testable.
- Preserves ease of use by providing a singleton instance for straightforward migration paths.
- Allows for more flexibility in various usage scenarios.

### Additional Information:

This change aligns with the modern best practices of Flutter development and makes the package more robust for complex applications.

### Acceptance Criteria:

- [x] All static methods are converted to instance methods.
- [x] Singleton instance is provided for backward compatibility.
- [x] Unit tests are added to verify that the changes meet the desired outcomes.

---

This PR aims to address [Issue #84](https://github.com/customerio/customerio-flutter/issues/84).

### DIFF
--- a/apps/amiapp_flutter/lib/src/app.dart
+++ b/apps/amiapp_flutter/lib/src/app.dart
@@ -78,7 +78,8 @@ class _AmiAppState extends State<AmiApp> {
             onLogin: (user) {
               _auth.login(user).then((signedIn) {
                 if (signedIn) {
-                  CustomerIO.identify(identifier: user.email, attributes: {
+                  CustomerIO.instance
+                      .identify(identifier: user.email, attributes: {
                     "first_name": user.displayName,
                     "email": user.email,
                     "is_guest": user.isGuest,
@@ -216,14 +217,14 @@ class _AmiAppState extends State<AmiApp> {
     if (_customerIOSDK.sdkConfig?.screenTrackingEnabled == true) {
       final Screen? screen = _router.location.toAppScreen();
       if (screen != null) {
-        CustomerIO.screen(name: screen.name);
+        CustomerIO.instance.screen(name: screen.name);
       }
     }
   }
 
   void _handleAuthStateChanged() {
     if (_auth.signedIn == false) {
-      CustomerIO.clearIdentify();
+      CustomerIO.instance.clearIdentify();
       _auth.clearUserState();
     }
   }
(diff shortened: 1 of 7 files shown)


Shahroz16 (changes requested):
Thank you for the PR @AristideVB, it indeed adds flexibility. But this is a breaking change as it would break the current implementation of the folks, so I would rather suggest deprecating the old methods and adding these new ones and then eventually removing the old methods later on.

AristideVB:
Thanks for the review, @Shahroz16. I get your point about avoiding breaking changes. However, I believe a version bump would be the most efficient approach here. It not only allows us to maintain a cleaner and more intuitive API but also avoids the issue of **naming conflicts**, which would be inevitable if we keep both the old and new methods. To make things smoother, I can provide a detailed migration guide for existing users. Let me know what you think.

Shahroz16:
@AristideVB Thank you again for your PR. I've shared it with the team and product, and we'll be reviewing it soon. I'll keep you updated on our next steps once I hear more.

Shahroz16:
The PR was very valuable, thank you for creating it. They will be bundled up with some of the other changes in the pipeline, that are going to affect the public-facing APIs.

Issue #87, opened 2023-09-24
Improve Screen Tracking by Implementing NavigatorObserver-Based CustomerIORouteObserver


AristideVB:
## Problem
The current approach to screen/route tracking relies on manual listener attachment for each route change in our app. The existing strategy is non-idiomatic for Flutter and potentially less efficient:

```dart
_router.addListener(() => _onRouteChanged());

void _onRouteChanged() {
  if (_customerIOSDK.sdkConfig?.screenTrackingEnabled == true) {
    final Screen? screen = _router.location.toAppScreen();
    if (screen != null) {
      CustomerIO.screen(name: screen.name);
    }
  }
}
```

## Proposed Solution
To solve this, I propose implementing `CustomerIORouteObserver` that leverages Flutter's `NavigatorObserver`. This approach aligns better with Flutter best practices and allows for more efficient and reliable route tracking. The GoRouter package supports route observers, allowing this new class to integrate seamlessly.

## Integration with GoRouter
GoRouter provides a way to attach observers to your routes, enabling cleaner and more idiomatic tracking of screen views. You can easily add the `CustomerIORouteObserver` to the GoRouter setup like this:

```dart
GoRouter(
  observers: [CustomerIORouteObserver()],
  // ... other configurations
)
```

By adding the observer directly into GoRouter's `observers` array, we can automatically and efficiently handle route tracking without requiring any manual intervention for each route change.

## Integration with MaterialApp
```dart
MaterialApp(
  home: MyAppHome(),
  navigatorObservers: [
    CustomerIORouteObserver(),
  ],
)
```
With this new setup, every time a route change occurs, the `CustomerIORouteObserver` will automatically fire and send the necessary tracking information to CustomerIO.

WellDunDun:
Great suggestions and I really hope they implement them! I've been really struggling to get screen tracking working with GoRouter and I'm in desperate need of some help.

aqrc:
There is such code for observer in official documentation which can be copy-pasted https://customer.io/docs/sdk/flutter/track-events/#auto-screenview

I agree though that it would be nice to have it out of the box)

jcpsimmons:
Closing as stale - this is from 2023 and screen tracking has evolved since. If the current NavigatorObserver behavior still falls short for you on the latest plugin, please reopen or open a fresh issue and we'll take a look.

PR #88, opened 2023-09-24
feat: add CustomerIORouteObserver for optimized route tracking

AristideVB:
## Summary

This PR introduces `CustomerIORouteObserver`, a route observer that automatically tracks screen views and reports them to Customer.io. The observer integrates seamlessly with GoRouter or MaterialApp, thus eliminating the need for manual tracking logic.

## Changes

- Implemented `CustomerIORouteObserver` that extends `NavigatorObserver`.
- Integrated the observer with GoRouter.
- Removed manual route tracking and listeners.

This PR aims to address [Issue #87](https://github.com/customerio/customerio-flutter/issues/87)

### DIFF
--- a/apps/amiapp_flutter/lib/src/app.dart
+++ b/apps/amiapp_flutter/lib/src/app.dart
@@ -1,4 +1,5 @@
 import 'package:customer_io/customer_io.dart';
+import 'package:customer_io/customer_io_observer.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:go_router/go_router.dart';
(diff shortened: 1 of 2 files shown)


Shahroz16:
The PR was very valuable, thank you for creating it. They will be bundled up with some of the other changes in the pipeline, that are going to affect the public-facing APIs.

AristideVB:
Thanks @Shahroz16 looking forward, for this update ? Why not merge my PR thought ? If you are going to use the code ? 🙂

_(Or suggest edits if needed)_

Shahroz16:
@AristideVB thank you again for the efforts here, really appreciate it. We are updating public APIs, and sample apps with the bigger change in question, we haven't decided yet, but we might expose a class for tracking which would be optional. 

Once we have all the details figured out, if we are to use this code, we will merge the PR as it is, otherwise, we will make related changes wherever needed.

PR #91, opened 2023-09-28
refactor: fix deployment 

Shahroz16:
inconsistency between pubignore and gitignore, also to make package lighter ignore sample app


apps/amiapp_flutter/ios/Flutter/flutter_export_environment.sh
mrehan27, line 1: Is this already added to `gitignore`? If not, we might need to add it there too

mrehan27 (approved):

apps/amiapp_flutter/ios/Flutter/flutter_export_environment.sh
levibostian: Also, I would expect that if this line was removed from the `.gitignore` this PR would have added this file in the diff?

Shahroz16: > Also, I would expect that if this line was removed from the .gitignore this PR would have added this file in the diff?
  
I would have hoped so as well, but i get this error 

```
 1 checked-in file is ignored by a `.gitignore`.
  Previous versions of Pub would include those in the published package.
  
  Consider adjusting your `.gitignore` files to not ignore those files, and if you do not wish to
  publish these files use `.pubignore`. See also dart.dev/go/pubignore
  
  Files that are checked in while gitignored:
  
  apps/amiapp_flutter/android/app/google-services.json
```

This makes me think that it was once uploaded in git but now its being ignored.

PR #99, opened 2023-11-08
ci: fix installing sd tool needed for deployment

levibostian:
The iOS SDK experienced a deployment failure on `main` because of a GH Action that we depend on. 

I installed a workaround in the iOS SDK until the GH action gets fixed. This PR applies the same workaround here.


Shahroz16 (approved):
Same comment as other PRs, approving to unblock PRs but lets make a follow up ticket to get a permanent solution for this.

PR #116, opened 2024-04-08
fix: added proguard rules for R8 strict mode

Shahroz16:
helps: [MBL-229 : Proguard rules for Android](https://linear.app/customerio/issue/MBL-229/proguard-rules-for-android)

What it does:
- Bumps the android SDK version
- Applied changes from[ Deprecated imperative apply of Flutter's Gradle plugins](https://docs.flutter.dev/release/breaking-changes/flutter-gradle-plugin-apply)

### DIFF
--- a/apps/amiapp_flutter/android/app/build.gradle
+++ b/apps/amiapp_flutter/android/app/build.gradle
@@ -1,20 +1,9 @@
-def localProperties = new Properties()
-def localPropertiesFile = rootProject.file('local.properties')
-if (localPropertiesFile.exists()) {
-    localPropertiesFile.withReader('UTF-8') { reader ->
-        localProperties.load(reader)
-    }
+plugins {
+    id "com.android.application"
+    id "kotlin-android"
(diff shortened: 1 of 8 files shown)


apps/amiapp_flutter/android/app/build.gradle
```diff
@@ -78,7 +68,7 @@ flutter {
 }
 
 dependencies {
-    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:1.7.21"
```
mrehan27, line 71: I think we should avoid hardcoding versions, as they are error-prone and harder to track when updating,

mrehan27 (approved):
LGTM 👍🏻

apps/amiapp_flutter/android/app/build.gradle
Shahroz16: agreed, left it for gradle 8 changes to not complicate this one.

Issue #113, opened 2024-03-20
Android Devices Not Registered in Production Mode


AristideVB:
### Issue Description
When using the Customer.io Flutter plugin, Android devices are correctly registered in debug mode but not in production. The same codebase registers devices on iOS in both debug and production modes without issues.

### Steps to Reproduce
1. Initialize Customer.io with the following configuration in the Flutter app:
   ```dart
   await CustomerIO.initialize(
     config: CustomerIOConfig(
       logLevel: kDebugMode ? CioLogLevel.debug : CioLogLevel.none,
       siteId: '[customerIOSiteId]',
       apiKey: '[customerIOApiKey]',
       region: Region.eu,
       enableInApp: true,
     ),
   );
   ```
2. Identify the user:
   ```dart
   CustomerIO.identify(identifier: user.email);
   ```
3. Call `CustomerIO.messagingPush()` to register the device for push notifications.

### Expected Behavior
Devices should be registered in both debug and production modes on Android, similar to the behavior observed on iOS devices.

### Actual Behavior
On Android, devices are only registered in debug mode. In production, the devices are not registered, leading to no push notifications being received.

### Additional Context
- This issue has been observed consistently across multiple Android devices.
- iOS devices do not exhibit this issue and are registered correctly in both debug and production modes.
- Relevant configurations and permissions have been double-checked for production builds.

### Environment
- **Customer.io Flutter Plugin Version**: 1.3.1
- **Flutter Version**: 3.19.3

AristideVB:
Note I've upgrade to Gradle 8, could it be related ?

& I have been overriding the namespace dynamically as suggested https://github.com/customerio/customerio-flutter/issues/107#issuecomment-1910507373

AristideVB:
Downgrading to Gradle 7.x fixed the issue in production, but we had to remove support for essential features in our app because of this downgrade...

Please upgrade the plugin to Gradle 8.x and having registerDevice work correctly on Android in production https://github.com/customerio/customerio-flutter/issues/107

levibostian:
Thank you for sharing this issue with us. Sorry for the frustration caused by this. 

I have added a +1 to our internal ticket for Gradle 8 support. Unfortunately, we do not have a release date for this support. Know that we are discussing it and will continue to prioritize it internally. 

Thank you again for sending this issue to us. Especially with all the very helpful details! 

I hope you have a great day.

mrehan27:
[...]

PR #117, opened 2024-04-09
feat: support for android gradle plugin 8

mrehan27:
closes: [MBL-232](https://linear.app/customerio/issue/MBL-232/flutter-sdk-compatibility-with-gradle-8)
blocked by: release of https://github.com/customerio/customerio-android/pull/296

### Highlights

- Support for Android Gradle Plugin 8.0
- JDK Version is updated to `17` from `1.8`

### Changes

- Updated flutter plugin to make it compatible with changes for Android Gradle Plugin 8.0 in native SDK
- Updated sample app to use same gradle plugin version

### Tasks

- [ ] Updated Android SDK version to actual expected value

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -9,7 +9,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:7.2.0'
+        classpath 'com.android.tools.build:gradle:8.3.1'
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
     }
 }
@@ -26,15 +26,16 @@ apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 
 android {
+    namespace 'io.customer.customer_io'
     compileSdkVersion 33
 
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_1_8
-        targetCompatibility JavaVersion.VERSION_1_8
+        sourceCompatibility JavaVersion.VERSION_17
+        targetCompatibility JavaVersion.VERSION_17
     }
 
     kotlinOptions {
-        jvmTarget = '1.8'
+        jvmTarget = '17'
     }
 
     sourceSets {
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -59,7 +59,7 @@ android {
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
     // Customer.io SDK
-    def cioVersion = "3.9.2"
+    def cioVersion = "3.10.0"
     implementation "io.customer.android:tracking:$cioVersion"
     implementation "io.customer.android:messaging-push-fcm:$cioVersion"
     implementation "io.customer.android:messaging-in-app:$cioVersion"
--- a/apps/amiapp_flutter/android/app/src/main/AndroidManifest.xml
+++ b/apps/amiapp_flutter/android/app/src/main/AndroidManifest.xml
@@ -1,5 +1,4 @@
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="io.customer.amiapp_flutter">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
 
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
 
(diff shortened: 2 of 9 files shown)


android/build.gradle
```diff
@@ -58,7 +59,7 @@ android {
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
     // Customer.io SDK
-    def cioVersion = "3.9.2"
+    def cioVersion = "3.10.0"
```
mrehan27, line 62: TODO: Make sure the version is correct here

apps/amiapp_flutter/android/app/src/main/AndroidManifest.xml
```diff
@@ -59,6 +59,15 @@
                 <data android:pathPrefix="/dashboard" />
             </intent-filter>
         </activity>
+        <!-- Register Customer.io Firebase Messaging Service as we want our SDK to handle all push
+             notifications -->
+        <service
+            android:name="io.customer.messagingpush.CustomerIOFirebaseMessagingService"
```
mrehan27, line 65: The notifications were not working for Android apps, so I added Option 2 so our SDK handles everything in it.

AristideVB: Great thanks, can you also make sure Android devices are well registered with this update ?

https://github.com/customerio/customerio-flutter/issues/113

Shahroz16 (approved):
Looks good. Adding the blocked label to wait until Android SDK gets merged.

apps/amiapp_flutter/android/app/src/main/AndroidManifest.xml
AristideVB: Hi @mrehan27 I've just tested [version 1.3.2](https://github.com/customerio/customerio-flutter/releases/tag/1.3.2) & it does seem to have resolved the issue https://github.com/customerio/customerio-flutter/issues/113 🙂

Thank you 🙏

mrehan27: Glad to hear. Thanks for confirming 👍🏻

apps/amiapp_flutter/android/app/src/main/AndroidManifest.xml
```diff
@@ -59,6 +59,18 @@
                 <data android:pathPrefix="/dashboard" />
             </intent-filter>
         </activity>
+        <!-- Register Customer.io Firebase Messaging Service as we want our SDK to handle all push
+             notifications. This is not required. However, if you have multiple services added for
+             handling push notifications, you might need to add this service to ensure all
+             notifications are handled by Customer.io. Read more:
+             https://www.customer.io/docs/sdk/flutter/push-notifications/multiple-push-providers/ -->
+        <service
```
levibostian, line 67: 🤔 The fact we need to make any change to our manifest makes me think that this is not a backwards compatible change. 

From [the comment below](https://github.com/customerio/customerio-flutter/pull/117/files#r1557529353), it sounds like you're not sure why this is needed but it has resolved your issue with push not showing up. 

I think we need to either figure out why this fixed the issue and make a change to our SDK so Flutter customers don't need to also make this change to their apps, or, we make a major version bump of our SDK and require a migration.

levibostian (changes requested):
Blocking because of manifest file conversation. I think that the PR as it is now is not a backwards compatible change.

apps/amiapp_flutter/android/app/src/main/AndroidManifest.xml
Shahroz16: Levi, in the last release, `firebase_messaging` was added in the sample application. When another messaging service is added, we already expect customers to take either of these options [as mentioned in the docs. ](https://customer.io/docs/sdk/flutter/push-notifications/multiple-push-providers/#cio-process-notification-payloads)

So we know why the push was not showing up. This comment might be improved, but the reason there is `might` is actually because of the uncertainty, which one of the `FirebaseServices` takes control over the messages, ours or another service adding theirs. Hence, expect to add these lines of code to ensure its always us.

mrehan27: Shahroz is correct. The issue is reproducible in `main` too. Also, the change is in sample app and is enforced due to multiple push providers not by our SDK or Gradle 8.

levibostian: Got it. 

Because this service was added in this PR made me think that gradle 8 support is what caused this. But because that's not correct, we're good to go. I'll resolve this and approve.

levibostian (approved):
Team responded back to my concerns on backwards compatibility. Approved!

PR #129, opened 2024-06-18
chore: test multiple push event handlers in sample app

levibostian:
Part of: https://linear.app/customerio/issue/MBL-383/[bug]-customer-unable-to-handle-push-notification-event-on-ios-with

Tests [this iOS bug fix](https://github.com/customerio/customerio-ios/pull/743) in FCM sample app, while also testing local push notifications in the app. 

This PR: 
- displays local push notifications, which we use for QA testing local push
- adds a push event handler for local push in `AppDelegate`. 

After this PR, the FCM sample app will have 3 push event handlers: CIO SDK, flutterfire, and the host iOS app (AppDelegate). This environment allows us to test the iOS bug that is caused when there are at least 2 push event handlers installed in an app. 

Testing:
* Local push QA regression tests. 
* Non-CIO FCM push QA regression tests (send a FCM push from Firebase console instead of Fly). 
For all QA tests, we expect that both flutterfire and the host app are able to handle the push events.

### DIFF
--- a/apps/amiapp_flutter/ios/Flutter/AppFrameworkInfo.plist
+++ b/apps/amiapp_flutter/ios/Flutter/AppFrameworkInfo.plist
@@ -21,6 +21,6 @@
   <key>CFBundleVersion</key>
   <string>1.0</string>
   <key>MinimumOSVersion</key>
-  <string>11.0</string>
+  <string>12.0</string>
 </dict>
 </plist>
--- a/apps/amiapp_flutter/ios/Podfile
+++ b/apps/amiapp_flutter/ios/Podfile
@@ -43,9 +43,9 @@ target 'Runner' do
   use_modular_headers!
   
   # Uncomment only 1 of the lines below to install a version of the iOS SDK
-  pod 'CustomerIO/MessagingPushFCM', '~> 2.11' # install production build
+  pod 'CustomerIO/MessagingPushFCM', '~> 2.13' # install production build
   # install_non_production_ios_sdk_local_path(local_path: '~/code/customerio-ios/', is_app_extension: false, push_service: "fcm")
-  install_non_production_ios_sdk_git_branch(branch_name: 'levi/v2-multiple-push-handlers', is_app_extension: false, push_service: "fcm")
+  # install_non_production_ios_sdk_git_branch(branch_name: 'levi/v2-multiple-push-handlers', is_app_extension: false, push_service: "fcm")
 
   flutter_install_all_ios_pods File.dirname(File.realpath(__FILE__))
 end
@@ -53,9 +53,9 @@ end
 target 'NotificationServiceExtension' do
   use_frameworks!
   # Uncomment only 1 of the lines below to install a version of the iOS SDK
-  pod 'CustomerIO/MessagingPushFCM', '~> 2.11' # install production build
+  pod 'CustomerIO/MessagingPushFCM', '~> 2.13' # install production build
   # install_non_production_ios_sdk_local_path(local_path: '~/code/customerio-ios/', is_app_extension: true, push_service: "fcm")
-  install_non_production_ios_sdk_git_branch(branch_name: 'levi/v2-multiple-push-handlers', is_app_extension: true, push_service: "fcm")
+  # install_non_production_ios_sdk_git_branch(branch_name: 'levi/v2-multiple-push-handlers', is_app_extension: true, push_service: "fcm")
 end
 
 post_install do |installer|
--- a/apps/amiapp_flutter/ios/Podfile
+++ b/apps/amiapp_flutter/ios/Podfile
@@ -1,7 +1,10 @@
-# Import Ruby functions from native SDK to more easily allow installing non-production SDK builds.
+# ------------- 
+# This code only used internally for Customer.io testing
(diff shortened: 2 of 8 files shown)


apps/amiapp_flutter/ios/Podfile
levibostian, line 1: Only here for testing until the native iOS SDK goes to prod.

apps/amiapp_flutter/ios/Runner/AppDelegate.swift
```diff
@@ -26,6 +26,8 @@ import FirebaseCore
             config.logLevel = .debug
         }
         MessagingPushFCM.initialize(configOptions: nil)
+        
+        UNUserNotificationCenter.current().delegate = self
```
levibostian, line 30: Registers the 3rd push event listener to the SDK.

apps/amiapp_flutter/ios/Runner/AppDelegate.swift
```diff
@@ -38,6 +40,39 @@ import FirebaseCore
     override func application(_ application: UIApplication, didFailToRegisterForRemoteNotificationsWithError error: Error) {
         MessagingPush.shared.application(application, didFailToRegisterForRemoteNotificationsWithError: error)
     }
+    
+    // Function called when a push notification is clicked or swiped away.
+    override func userNotificationCenter(_ center: UNUserNotificationCenter, didReceive response: UNNotificationResponse, withCompletionHandler completionHandler: @escaping () -> Void) {
```
Shahroz16, line 45: We don't have the support to observe these callbacks in the `flutter/dart` file? similar to what we are doing in react native?

levibostian: Good question. There probably is a way, but I was not able to find a way to do it. I am guessing it will require a flutter plugin to do that? From searching, I have not been able to find one. Happy to change it if someone finds a resource.

Shahroz16: You have already added the plugin "flutter_local_notifications", [here](https://blog.codemagic.io/flutter-local-notifications/) is something I found.

Shahroz16 (commented):
The comment didn't show up earlier, 
 
> You have already added the plugin "flutter_local_notifications", [here](https://blog.codemagic.io/flutter-local-notifications/) is something I found.

apps/amiapp_flutter/ios/Runner/AppDelegate.swift
levibostian: Oh, I forgot to mention that I did try to install a callback for `flutter_local_notifications` but was not able to successfully get it to work. I installed the callback functions according to the official docs, I noticed that the CIO SDK successfully installed the native push event handler for the plugin, but my callback function never got executed. 

Rather then dig deeper into it, I decided to go with this approach since I knew it would work. 

I have not found the docs for the plugin to be very helpful so it would take me a bit of time to debug this and see why the callback function is not being called.

apps/amiapp_flutter/ios/Flutter/AppFrameworkInfo.plist
levibostian, line 1: Flutter updated this file automatically. Happy to move to another PR if requested.

apps/amiapp_flutter/ios/Podfile
levibostian, line 1: Fixed script to install non-production builds. Script is now located in v2 branch instead of main.

apps/amiapp_flutter/ios/Runner/AppDelegate.swift
levibostian: Got Dart callbacks working with `flutter_local_notifications` and pushed the commit to this PR. 

I figured out the problem. I was overriding the native push event callback functions in the `AppDelegate` which then prevented `flutter_local_notifications` from ever getting called. 

You'll now see in the PR that there is only 1 line required in `AppDelegate` to get this to work.

apps/amiapp_flutter/ios/Podfile
```diff
@@ -1,7 +1,10 @@
-# Import Ruby functions from native SDK to more easily allow installing non-production SDK builds.
+# ------------- 
+# This code only used internally for Customer.io testing
 require 'open-uri'
-IO.copy_stream(URI.open('https://raw.githubusercontent.com/customerio/customerio-ios/main/scripts/cocoapods_override_sdk.rb'), "/tmp/override_cio_sdk.rb")
+IO.copy_stream(URI.open('https://raw.githubusercontent.com/customerio/customerio-ios/v2/scripts/cocoapods_override_sdk.rb'), "/tmp/override_cio_sdk.rb")
```
Shahroz16, line 4: do we still need this, now that we are pointing towards production?

Shahroz16 (approved):
Thank you for the changes 🙌

apps/amiapp_flutter/ios/Podfile
levibostian: I think you're asking if we should remove this block of code that installs the script? 

This PR fixes the script that was already in the `Podfile`. So I didn't add anything new to this file.

PR #133, opened 2024-07-30
build: get sample apps compiling after flutter update

levibostian:
When trying to get the flutter sample app compiling on iOS, I got some errors during compilation. These updates were what I needed to change to get the app to compile again.

### DIFF
--- a/apps/amiapp_flutter/lib/src/color_schemes.g.dart
+++ b/apps/amiapp_flutter/lib/src/color_schemes.g.dart
@@ -27,7 +27,6 @@ const lightColorScheme = ColorScheme(
   onErrorContainer: Color(0xFF410002),
   surface: Color(0xFFFFFBFF),
   onSurface: Color(0xFF1B1B1F),
-  surfaceContainerHighest: Color(0xFFE3E1EC),
   onSurfaceVariant: Color(0xFF46464F),
   outline: Color(0xFF777680),
   onInverseSurface: Color(0xFFF3F0F4),
@@ -37,6 +36,8 @@ const lightColorScheme = ColorScheme(
   surfaceTint: Color(0xFF4D57A9),
   outlineVariant: Color(0xFFC7C5D0),
   scrim: Color(0xFF000000),
+  background: Color(0xFFF3F0F4),
+    onBackground: Color(0xFF000000)
 );
 
 const darkColorScheme = ColorScheme(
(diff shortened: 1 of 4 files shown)


apps/amiapp_flutter/lib/src/color_schemes.g.dart
levibostian, line 1: The colors I picked were somewhat random. I chose white colors for the light theme and dark colors for the dark theme.

apps/amiapp_flutter/lib/src/color_schemes.g.dart
```diff
@@ -27,7 +27,6 @@ const lightColorScheme = ColorScheme(
   onErrorContainer: Color(0xFF410002),
   surface: Color(0xFFFFFBFF),
   onSurface: Color(0xFF1B1B1F),
-  surfaceContainerHighest: Color(0xFFE3E1EC),
```
levibostian, line 30: Compiler said that this property is no longer used.

apps/amiapp_flutter/lib/src/color_schemes.g.dart
```diff
@@ -69,4 +69,6 @@ const darkColorScheme = ColorScheme(
   surfaceTint: Color(0xFFBCC2FF),
   outlineVariant: Color(0xFF46464F),
   scrim: Color(0xFF000000),
+    background: Color(0xFF000000),
+    onBackground: Color(0xFFC7C5D0)
```
mrehan27, line 73: I think the file needs a few indentation fixes

apps/amiapp_flutter/pubspec.lock
```diff
@@ -678,4 +678,4 @@ packages:
     version: "3.1.2"
 sdks:
   dart: ">=3.3.0 <4.0.0"
-  flutter: ">=3.18.0-18.0.pre.54"
+  flutter: ">=3.7.0"
```
mrehan27, line 681: I see this and a few other changes for downgrading versions. Could this be the reason for getting those errors? Can you try updating Flutter version and see if that helps?

If that fixes the problem, would be happy to discuss internally on how can we make this consistent across our team and have clear recommendations for versions. But I want to make sure we understand the root cause and we have a reason to downgrade these versions.

mrehan27 (commented):
My only concern is around downgrading versions. Once that discussion is resolved, the PR looks good to merge.

apps/amiapp_flutter/pubspec.lock
levibostian: Oh, that is weird. I am not sure why this happened? 

Let me try an upgrade first and see what happens.

levibostian: Not sure what caused these lock file changes but I seem to have fixed it. 

The iOS app is compiling successfully with the code in this PR. Ready for re-review.

mrehan27 (approved):
Looks good 👍🏻

levibostian:
I think that my version of flutter was out-of-date or missconfigured on my machine. 

I upgraded to the latest version of Flutter and now the app runs on `main` with no modifications. Closing PR, ignore it.

PR #136, opened 2024-08-09
chore: sample app fixes

mrehan27:

### Changes

- Fixed sample app so it can compile
- Updated `build.gradle` as required by `flutter_local_notifications`
- Updated pub dependencies to reduce challenges on CI builds


apps/amiapp_flutter/android/app/build.gradle
```diff
@@ -40,6 +43,7 @@ android {
         targetSdkVersion 33
         versionCode 1
         versionName "1.0"
+        multiDexEnabled true
```
levibostian, line 46: I think that this increases build times when it's enabled so just calling it out to make sure this is required and it wasn't automatically added or you forgot to remove it. If you added it intentionally, no problem.

apps/amiapp_flutter/lib/src/utils/extensions.dart
```diff
@@ -99,3 +100,16 @@ extension AmiAppDoubleExtensions on double {
     return toStringAsFixed(0);
   }
 }
+
+extension LocationExtensions on GoRouter {
+  // Get location of current route
+  // This is a workaround to get the current location as location property
```
levibostian, line 106: Since this is labeled as a workaround, it makes me wonder if this is a short-term solution? To make our sample app an up-to-date reference for customers, should we replace this with recommended best practice code in the future?

levibostian (approved):

apps/amiapp_flutter/android/app/build.gradle
mrehan27: Correct. This too was added only for [flutter_local_notifications gradle setup](https://pub.dev/packages/flutter_local_notifications#gradle-setup).

apps/amiapp_flutter/lib/src/utils/extensions.dart
mrehan27: Agreed. I'm not expert with `GoRouter` so not fully sure about best practices for it at the moment. But yes, we can put at a ticket to simplify and improve this for customers. Sample apps project could be the best time to handle this? 🤔

apps/amiapp_flutter/android/app/build.gradle
```diff
@@ -7,10 +7,13 @@ plugins {
 
 android {
     namespace 'io.customer.amiapp_flutter'
-    compileSdkVersion 33
+    compileSdkVersion 34
     ndkVersion flutter.ndkVersion
 
     compileOptions {
+        // Flag to enable support for the new language APIs
+        coreLibraryDesugaringEnabled true
```
Shahroz16, line 15: interesting, what new language API are we introducing?

Shahroz16 (approved):

apps/amiapp_flutter/android/app/build.gradle
mrehan27: This is required by [flutter_local_notifications gradle setup](https://pub.dev/packages/flutter_local_notifications#gradle-setup).

PR #137, opened 2024-08-09
chore: fastlane updates to build sample app

mrehan27:

### Changes

- Updated `Gemfile` and `Gemfile.lock` to include required dependencies
- Updated `Appfile`, `Gymfile`, `Matchfile` and `Pluginfile` to match values required by Fastlane to build sample apps on CI
- Broke down `Fastfile` into following smaller files
  - `build_helper.rb`-> Helps building and uploading Android and iOS builds
  - `github_helper.rb` -> Helps fetching Github attributes conveniently (mostly code movement only)
  - `version_helper.rb` -> Helps updating app and SDK versions on CI


apps/amiapp_flutter/fastlane/Fastfile
```diff
@@ -1,341 +1,2 @@
...
-        key_content: ENV["APP_STORE_CONNECT_API_KEY_CONTENT_B64"],
-        is_key_content_base64: true,
-        in_house: is_enterprise_app
-      )
-    end
-  end
-
-  # example for main builds: `bundle exec fastlane android deploy_app version:1.0.0"`
-  # example for develoment builds (pull request, push): `bundle exec fastlane deploy_app`  
-  lane :deploy_app do |values|
```
mrehan27, line 108: Some of this is removed and some moved and updated in `build_helper.rb` and `version_helper.rb`

apps/fastlane/Fastfile
```diff
@@ -0,0 +1,48 @@
...
+  pubspec_content = File.read(pubspec_path)
+  new_pubspec_content = pubspec_content.gsub(/^version: .*/, "version: #{new_version}")
+
+  File.write(pubspec_path, new_pubspec_content)
+  UI.message("Updated version to #{new_version} in #{pubspec_path}")
+end
+
+# Lane to update Flutter app version
+lane :update_flutter_sdk_version do |options|
+  project_path = options[:project_path] || File.join(Dir.pwd, '../../..')
```
levibostian, line 28: Is `File.join(Dir.pwd, '../../..')` safe to use, or do you think it will result in bugs if someone moves a file and something breaks? 

I am OK with the code, but I wonder if maybe a `UI.message(` right above would help if it does break, we can fix it quickly. 

```suggestion
  project_path = options[:project_path] || File.join(Dir.pwd, '../../..')
  UI.message("updating version for files in path....")  
```

apps/fastlane/Fastfile
```diff
@@ -0,0 +1,48 @@
...
+  version_name = options[:version_name] || ENV['SDK_VERSION_NAME']
+
+  UI.message("Updating sdk version to #{version_name} in #{project_path}")
+  update_pubspec_version(project_path, version_name)
+end
+
+# Lane to update Flutter app version
+lane :update_flutter_app_version do |options|
+  project_path = options[:project_path] || File.join(Dir.pwd, '..')
```
levibostian, line 37: Same question as above with use of `File.join(Dir.pwd, '..')`

apps/fastlane/helpers/build_helper.rb
```diff
@@ -0,0 +1,132 @@
...
+require 'xcodeproj'
+require_relative 'github_helper.rb'
+
+platform :android do
+  lane :build do |values|
+    build_notes = get_build_notes()
+    test_groups = get_build_test_groups()
+    app_package_name = CredentialsManager::AppfileConfig.try_fetch_value(:package_name) # get package_name from Appfile
+
+    UI.important(find_firebase_app_id(app_identifier: app_package_name))
```
levibostian, line 11:
```suggestion
    UI.important("Firebase app id: #{find_firebase_app_id(app_identifier: app_package_name)}"))
```

This code doesn't compile, but trying to illustrate a suggestion of labeling what the value is so it's easier to read in CI logs.

apps/fastlane/helpers/build_helper.rb
```diff
@@ -0,0 +1,132 @@
...
+
+    # Build release APK using Flutter CLI
+    sh("flutter build apk --release")
```
levibostian, line 14: Do you suggest we build an APK compared to AAB?

apps/fastlane/helpers/build_helper.rb
levibostian: I like the comment you made later on, `# Firebase app id is required. Get it from google-services.json file`. perhaps we can include that here, too?

levibostian (approved):

levibostian:
Great job!

apps/fastlane/Fastfile
mrehan27: Yup, it can surely break if the file is moved. To help with this, there is [already a message](https://github.com/customerio/customerio-flutter/pull/137/files#diff-d67e154c29ddfa99f01998d885b48404fa005f335d4efee735ef298efafa314dR31) just below this line:

```
UI.message("Updating sdk version to #{version_name} in #{project_path}")
```

Are you suggesting adding another message, or was it just missed?

apps/fastlane/Fastfile
mrehan27: There is [a message](https://github.com/customerio/customerio-flutter/pull/137/files#diff-d67e154c29ddfa99f01998d885b48404fa005f335d4efee735ef298efafa314dR44) for this one too

```
UI.message("Updating app versions to #{new_version} in #{project_path}")
```

apps/fastlane/helpers/build_helper.rb
mrehan27: I did consider using AAB as well, but then switched to APK because I think it will be easier for anyone to download from Firebase and install the APK if needed, especially for non-registered devices (e.g., emulators).

Shahroz16 (approved):

PR #138, opened 2024-08-09
ci: build sample app

mrehan27:

### Changes

- Updated github actions to build sample apps for every pull request created


.github/actions/setup-android/action.yml
mrehan27, line 1: I think we can reuse these actions in our repos, but I think would be easier if we keep do it after CI is working on all repos

levibostian: You sure can re-use 🙂

Given the setup-action is in the Android repo, from flutter SDK GH workspaces files: 

```yml
uses: customerio/customerio-android/.github/actions/setup-android@main
```

.github/workflows/build-sample-apps.yml
```diff
@@ -0,0 +1,195 @@
...
+          lane: 'android build'
+        env:
+          FIREBASE_APP_DISTRIBUTION_SERVICE_ACCOUNT_CREDS_B64: ${{ secrets.FIREBASE_APP_DISTRIBUTION_SERVICE_ACCOUNT_CREDS_B64 }}
+
+      - name: Setup iOS environment for sample app
+        uses: ./.github/actions/setup-ios
+        with:
+          xcode-version: ${{ env.XCODE_VERSION }}
+
+      - name: Cache CocoaPods downloaded dependencies for faster builds in the future
```
levibostian, line 156: instead of compiling android and ios apps inside of the same job, I suggest separating them into their own `job` so they can run in parallel. Could speedup the CI running.

levibostian (approved):

.github/actions/setup-android/action.yml
mrehan27: Thanks, updated this and removed copied actions.

.github/actions/setup-ios/action.yml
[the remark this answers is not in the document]
mrehan27: Agreed. Updated this. The only change we should make is to not hardcode any versions like XCode Version, Ruby Version, etc. So in case we need to modify it, we can pass them optionally. For now, same version seems to work well.

.github/workflows/build-sample-apps.yml
```diff
@@ -0,0 +1,195 @@
...
+        run: echo "RUBY_VERSION=$(cat .ruby-version)" >> $GITHUB_ENV
+
+      - name: Install Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ env.RUBY_VERSION }}
+          bundler-cache: true # cache tools to make builds faster in future
+
+      - name: Install Ruby dependencies
+        run: bundle install
```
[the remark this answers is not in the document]
mrehan27: Unfortunately it [seems to fail](https://github.com/customerio/customerio-flutter/actions/runs/10367890024/job/28700314375), so I have reverted the change.

.github/workflows/build-sample-apps.yml
mrehan27: This is a good idea as this will not only speed up the process but also prevent the other app not to fail if one of them is not compiling. But the changes might be larger, I'll do them in another PR later after trying it out for React Native apps first.

.github/workflows/build-sample-apps.yml
```diff
@@ -0,0 +1,195 @@
+name: Build sample apps
+
+on:
+  pull_request: # build sample apps for every commit pushed to an open pull request (including drafts)
+  push:
+    branches: [ main, feature/* ]
+  release: # build sample apps for every git tag created. These are known as "stable" builds that are suitable for people outside the mobile team.
+    types: [ published ]
```
Shahroz16, line 8: Just a note, `published` has previously not worked for us, especially when kicking off the build release workflow. Its been very flaky and we have to update the CD

Shahroz16 (commented):
looks good as long as we can verify the `release` as a trigger

.github/workflows/build-sample-apps.yml
mrehan27: Removed

.github/workflows/build-sample-apps.yml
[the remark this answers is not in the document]
levibostian: Not a big deal if we need to run `bundle install` a 2nd time.

the `setup-ruby` action [does come with an option](https://github.com/ruby/setup-ruby?tab=readme-ov-file#working-directory) to set a directory where the `Gemfile.lock` is. If you set the `setup-ruby` directory to the app's subdirectory, I wonder if that would fix it? 

Again, not a big deal if we have this duplication.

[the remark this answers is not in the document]
mrehan27: Yes, it worked. Thanks 👍🏻

PR #152, opened 2024-10-28
chore: update public API

Shahroz16:
closes: https://linear.app/customerio/issue/MBL-618/implement-public-api-structure-for-instance-pattern

changes:
- updated public API to be more testable and use dart/flutter conventions
- updated the tests and sample app to utilize the updated APIs


lib/customer_io.dart
```diff
@@ -8,23 +10,70 @@ import 'messaging_in_app/platform_interface.dart';
...
+  CustomerIOMessagingInAppPlatform get inAppMessaging => _inAppMessaging;
 
   /// To initialize the plugin
   ///
   /// @param config includes required and optional configs etc
-  static Future<void> initialize({
-    required CustomerIOConfig config,
-  }) {
-    return _customerIO.initialize(config: config);
+  static Future<void> initialize({required CustomerIOConfig config}) async {
```
Shahroz16, line 70: this is still static function, because `CustomerIO.instance.initialize` didn't look good and then I checked what others could be doing, and saw Firebase also has initialize method as static while others instance bound.

mrehan27 (approved):
Looks good 👍🏻

lib/customer_io.dart
```diff
@@ -8,23 +10,73 @@ import 'messaging_in_app/platform_interface.dart';
...
+    // Check if already initialized
+    if (_instance == null) {
+      // Create new instance if not initialized
+      _instance = CustomerIO._();
+      // Initialize the platform
+      await _instance!._platform.initialize(config: config);
+    } else {
+      print('CustomerIO SDK has already been initialized');
```
mrehan27, line 78: Can we please update this to fix lint?

test/customer_io_test.dart
mrehan27, line 1: The tests are also failing, can you please fix them before merging?

mrehan27 (approved):
I just noticed a few CI checks failing, we just need to fix them before merging. Rest looks good.

PR #153, opened 2024-10-29
chore: update iOS native dependencies to Data Pipelines

mrehan27:
part of [MBL-621](https://linear.app/customerio/issue/MBL-621/update-native-sdk-dependencies-to-data-pipelines-release)

### Changes

- Updated iOS native SDK dependencies from Tracking to Data Pipelines
- Removed outdated imports and commented legacy SDK calls to allow stepwise integration of new implementation
- Added TODOs to each commented legacy SDK call to ensure nothing is overlooked
- Fixed sample app compatibility issues on newer Xcode versions


apps/amiapp_flutter/ios/Runner/AppDelegate.swift
```diff
@@ -21,11 +20,17 @@ import FirebaseCore
 
         Messaging.messaging().delegate = self
         
+        // TODO: Fix SDK initialization
```
Shahroz16, line 23: should we not just remove it? because its not going to be needed anymore?

Shahroz16 (approved):

apps/amiapp_flutter/ios/Runner/AppDelegate.swift
mrehan27: Yeah, you're right. I initially kept it thinking it would be replaced with `MessagingPush` initialization. But since it's already in a separate block, we can remove it entirely. I've removed it from the code.

PR #160, opened 2024-11-06
chore: initialize native ios sdk in flutter

mrehan27:
part of [MBL-620](https://linear.app/customerio/issue/MBL-620/initialization)

### Changes

- Initialized native iOS SDK using configuration provided by Flutter
- Added `SDKConfigBuilder` mapper class for simplified config parsing


Shahroz16 (approved):
looks good

PR #162, opened 2024-11-07
chore: update sample app settings ui for data pipelines

mrehan27:
part of [MBL-623](https://linear.app/customerio/issue/MBL-623/update-sample-app-settings-screen)

### Changes

- Adjusted settings UI in sample app to align with Data Pipelines configuration
- Modified SDK initialization to use stored values
- Updated input requirements to make non-required fields optional

### DIFF
--- a/apps/amiapp_flutter/lib/src/utils/extensions.dart
+++ b/apps/amiapp_flutter/lib/src/utils/extensions.dart
@@ -32,6 +32,10 @@ extension AmiAppExtensions on BuildContext {
 extension AmiAppStringExtensions on String {
   bool equalsIgnoreCase(String? other) => toLowerCase() == other?.toLowerCase();
 
+  String? nullIfEmpty() {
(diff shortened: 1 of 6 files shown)


apps/amiapp_flutter/lib/src/utils/extensions.dart
```diff
@@ -92,7 +108,7 @@ extension AmiAppStringExtensions on String {
   }
 }
 
-extension AmiAppDoubleExtensions on double {
+extension AmiAppDoubleExtensions on int {
```
Shahroz16, line 111: double extension on int? 🤔

Shahroz16 (approved):

apps/amiapp_flutter/lib/src/utils/extensions.dart
mrehan27: Fixed it 😅

PR #166, opened 2024-11-08
ci: ensure task fails if any build fails

mrehan27:
### Changes

- Modified build sample apps task to fail if either Android or iOS build fails
- Updated `version_helper.rb` to align with React Native and prevent failure when time starts with `0`

### Notes

- Created PR as I observed an issue in Data Pipelines support changes where tasks didn't fail on individual build failures, leading to false positives.
- Once merged to `main`, I'll merge this change to the Data Pipelines feature branch.

### DIFF
--- a/apps/fastlane/helpers/version_helper.rb
+++ b/apps/fastlane/helpers/version_helper.rb
@@ -13,14 +13,30 @@ lane :generate_new_version do |options|
 
   # Replace '/' with '-' to avoid issues with unsupported characters in version name
   branch_name = branch_name.gsub('/', '-')
-  timestamp = current_time.strftime("%Y%m%d.%H%M%S")
+  # Extract ticket number from branch name
+  ticket_number_in_branch_name = branch_name.scan(/\d+/).join
+  # If no ticket number found, set it to 0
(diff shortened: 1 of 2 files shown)


apps/fastlane/helpers/version_helper.rb
```diff
@@ -13,14 +13,30 @@
...
+  hour = current_time.hour.to_s
+  minute = current_time.min.to_s
+  second = current_time.sec.to_s
+
+  # Combine them into the desired format
+  major = "#{year}#{month}#{day}"
+  minor = "#{hour}#{minute}#{second}"
+  patch = ticket_number_in_branch_name.to_s
+
+  sdk_version_name = "#{major}.#{minor}.#{patch}"
```
mrehan27, line 34: This matches with that we do in [React Native](https://github.com/customerio/customerio-reactnative/blob/e8535a10c0e2d1e07fb2b160cc1d194e54c6e2f3/Apps/fastlane/helpers/version_helper.rb#L5-L48) today

Shahroz16: not fully sure whats happening here and why its needed, can you add some comments? i understand its the same in react native and we can probably add comments there too. 😕

Shahroz16 (approved):

PR #167, opened 2024-11-08
chore: update screen tracking parameters

mrehan27:
closes: [MBL-634](https://linear.app/customerio/issue/MBL-634/screen-events)

### Changes

- Updated SDK `screen` method to use revised parameter names
- Adjusted native SDK, tests, and sample app accordingly

#### Before

```dart
CustomerIO.screen(name: "Favourites");
CustomerIO.screen(name: "Settings", attributes: {
  "source": "Dashboard",
  "is_logged_in": true,
});
```

#### After

```dart
CustomerIO.instance.screen(title: "Favourites");
CustomerIO.instance.screen(title: "Settings", properties: {
  "source": "Dashboard",
  "is_logged_in": true,
});
```

### DIFF
--- a/apps/amiapp_flutter/lib/src/data/config.dart
+++ b/apps/amiapp_flutter/lib/src/data/config.dart
@@ -8,8 +8,8 @@ class CustomerIOSDKConfig {
   final String cdpApiKey;
   final String? migrationSiteId;
   final Region? region;
-  final bool? debugModeEnabled;
-  final bool? screenTrackingEnabled;
+  final bool debugModeEnabled;
+  final bool screenTrackingEnabled;
   final bool? autoTrackDeviceAttributes;
   final String? apiHost;
   final String? cdnHost;
@@ -22,8 +22,8 @@ class CustomerIOSDKConfig {
     required this.cdpApiKey,
     this.migrationSiteId,
     this.region,
-    this.debugModeEnabled,
-    this.screenTrackingEnabled,
+    this.debugModeEnabled = true,
+    this.screenTrackingEnabled = true,
     this.autoTrackDeviceAttributes,
     this.apiHost,
     this.cdnHost,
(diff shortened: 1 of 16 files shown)


apps/amiapp_flutter/lib/src/data/config.dart
```diff
@@ -8,8 +8,8 @@ class CustomerIOSDKConfig {
   final String cdpApiKey;
   final String? migrationSiteId;
   final Region? region;
-  final bool? debugModeEnabled;
-  final bool? screenTrackingEnabled;
+  final bool debugModeEnabled;
+  final bool screenTrackingEnabled;
```
mrehan27, line 12: Because these are only app settings, I updated them to non-nullable

apps/amiapp_flutter/lib/src/utils/extensions.dart
```diff
@@ -108,15 +108,6 @@ extension AmiAppStringExtensions on String {
   }
 }
 
-extension AmiAppIntExtensions on int {
-  String? toTrimmedString() {
```
mrehan27, line 112: Since we now use Int for flush properties, we no longer need this extension

lib/customer_io.dart
```diff
@@ -88,9 +89,9 @@ class CustomerIO {
   /// @param userId unique identifier for a profile
   /// @param traits (Optional) params to set profile attributes
   void identify(
-      {required String userId,
-      Map<String, dynamic> traits = const {}}) {
-    return _platform.identify(userId: userId, traits: traits);
+      {required String userId, Map<String, dynamic> traits = const {}}) {
+    return _platform.identify(
+        userId: userId, traits: traits.excludeNullValues());
```
mrehan27, line 94: Excluded `null` from all currently implemented functions, as we don't currently support `null` attributes in native SDKs. Open to any suggestions if there are other ideas on how to handle this!

lib/extensions/map_extensions.dart
```diff
@@ -0,0 +1,7 @@
+/// Extensions for [Map] class that provide additional functionality and convenience methods.
+extension CustomerIOMapExtension on Map<String, dynamic> {
+  /// Returns a new map with entries that have non-null values, excluding null values.
+  Map<String, dynamic> excludeNullValues() {
+    return Map.fromEntries(entries.where((entry) => entry.value != null));
```
Shahroz16, line 5:
```suggestion
    return  map.removeWhere((key, value) => value == null);
```
not this exact place but dart provides this utility method

Shahroz16 (approved):

lib/extensions/map_extensions.dart
mrehan27: It definitely sounds good, but I deliberately avoided modifying the same map because removing items directly could unintentionally alter customer provided map, which might cause issues in their app. So I created a new map instance to avoid any potential side effects.

PR #173, opened 2024-11-18
chore: added device attributes support

Shahroz16:
closes: [MBL-637: Device Attributes](https://linear.app/customerio/issue/MBL-637/device-attributes)

### DIFF
--- a/lib/customer_io_method_channel.dart
+++ b/lib/customer_io_method_channel.dart
@@ -182,7 +182,7 @@ class CustomerIOMethodChannel extends CustomerIOPlatform {
   @override
   void setDeviceAttributes({required Map<String, dynamic> attributes}) {
     try {
-      final payload = {TrackingConsts.attributes: attributes};
+      final payload = {TrackingConsts.traits: attributes};
       methodChannel.invokeMethod(MethodConsts.setDeviceAttributes, payload);
     } on PlatformException catch (exception) {
       handleException(exception);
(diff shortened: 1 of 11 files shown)


lib/customer_io_method_channel.dart
```diff
@@ -182,7 +182,7 @@ class CustomerIOMethodChannel extends CustomerIOPlatform {
   @override
   void setDeviceAttributes({required Map<String, dynamic> attributes}) {
     try {
-      final payload = {TrackingConsts.attributes: attributes};
+      final payload = {TrackingConsts.traits: attributes};
```
mrehan27, line 185: I think we should match this with either wrapper public API or native implementation. Since we don't use `traits` for devices in either, it would be better to stick with `attributes` for consistency.

mrehan27: We refer to profile attributes as `traits` in our `identify` method within Data Pipelines support, so it's not that confusing and is easier to relate to for anyone familiar with the SDK. However, we haven't referred to device attributes as `traits` anywhere so far and have only used `traits` specifically for profiles.

I'm okay with using either `attributes` or `traits` for profiles, but for devices, I feel using `traits` could be confusing. If we want to use `attributes` for profiles attributes as well, I'm completely fine with that since it matches the public API and is very clear.

Shahroz16: Moving all to `attributes` for consistency, if we want to map public API with params.

mrehan27 (approved):

PR #179, opened 2024-11-22
chore: rename android classes and method extensions

mrehan27:
part of: [MBL-657](https://linear.app/customerio/issue/MBL-657/refactor-and-clean-up-android-code-in-flutter-sdk)

### Changes

- Moved extensions to dedicated files for better organization
- Renamed `getAsTypeOrNull` to `getAs` for simplicity while retaining functionality
- Renamed `CustomerIoPlugin` to `CustomerIOPlugin` to align with naming conventions
- Renamed `CustomerIOPluginModule` to `NativeModuleBridge` for clarity, as it's an internal class and `CustomerIO` prefix looked redundant

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -39,7 +39,7 @@ flutter:
     platforms:
       android:
         package: io.customer.customer_io
-        pluginClass: CustomerIoPlugin
+        pluginClass: CustomerIOPlugin
       ios:
         pluginClass: CustomerIoPlugin
         native_sdk_version: 3.5.1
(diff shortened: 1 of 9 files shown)


pubspec.yaml
```diff
@@ -39,7 +39,7 @@ flutter:
     platforms:
       android:
         package: io.customer.customer_io
```
mrehan27, line 41: I think we can rename package to follow more conventional format, but I avoided doing it for now. Happy to make the change if we feel we should go for it.

Shahroz16 (approved):

PR #183, opened 2024-11-26
chore: refactor dart call forwarding

mrehan27:
part of: [MBL-660](https://linear.app/customerio/issue/MBL-660/refactor-and-clean-up-dart-code-in-flutter-sdk)

### Changes

- Relocated constants to local directories for each module
- Simplified forwarding calls to native methods by adding extensions
- Made `inAppMessaging` a static property to align with the initial proposal
- Updated public API to expose all methods as `Future`, reflecting asynchronous nature of native calls
- Fixed and updated tests to align with these changes


lib/customer_io.dart
```diff
@@ -61,17 +61,13 @@ class CustomerIO {
...
-            'Call CustomerIO.initialize() first.',
-      );
-    }
-    return _instance!._pushMessaging;
+    return instance._pushMessaging;
   }
 
   /// Access in-app messaging functionality
-  CustomerIOMessagingInAppPlatform get inAppMessaging => _inAppMessaging;
+  static CustomerIOMessagingInAppPlatform get inAppMessaging {
```
mrehan27, line 68: Updated to match with initial proposal and with `pushMessaging`

lib/customer_io_method_channel.dart
mrehan27, line 1: The file has been moved to `lib/data_pipelines/customer_io_method_channel.dart` below. But due to significant content changes (because of extension), GitHub shows it as a deleted file.

lib/data_pipelines/customer_io_method_channel.dart
```diff
@@ -0,0 +1,108 @@
...
+import 'package:flutter/services.dart';
+
+import '../customer_io_config.dart';
+import '../customer_io_enums.dart';
+import '../extensions/method_channel_extensions.dart';
+import '_native_constants.dart';
+import 'customer_io_platform_interface.dart';
+
+/// An implementation of [CustomerIOPlatform] that uses method channels.
+class CustomerIOMethodChannel extends CustomerIOPlatform {
```
mrehan27, line 13: Updated implementation of `lib/customer_io_method_channel.dart` using extensions

lib/customer_io.dart
```diff
@@ -95,7 +91,7 @@ class CustomerIO {
   ///
   /// @param userId unique identifier for a profile
   /// @param traits (Optional) params to set profile attributes
-  void identify(
+  Future<void> identify(
```
Shahroz16, line 94: not sure we need `void` here, its a fire and forget call.

lib/customer_io.dart
```diff
@@ -105,23 +101,23 @@ class CustomerIO {
...
+    return _platform.clearIdentify();
   }
 
   /// To track user events like loggedIn, addedItemToCart etc.
   /// You may also track events with additional yet optional data.
   ///
   /// @param name event name to be tracked
   /// @param properties (Optional) params to be sent with event
-  void track(
+  Future<void> track(
```
Shahroz16, line 113: same for this and others.

Shahroz16 (commented):
Looks good, we dont need to make all methods `Future` its only for those where we need to make sure native code is executed before the dart code, e.g `initialize` for rest we are okay to use just void.

mrehan27:
Discussed on call, and we decided to stick with `void`. I'll go ahead and revert all updates for `Future<void>`.

Shahroz16 (approved):

PR #186, opened 2024-11-29
chore: rename event listener for in-app

Shahroz16:


test/customer_io_test.mocks.dart
```diff
@@ -158,19 +146,4 @@ class MockTestCustomerIoPlatform extends _i1.Mock
         ),
         returnValueForMissingStub: null,
       );
-  _i2.StreamSubscription<dynamic> subscribeToInAppEventListener(
```
Shahroz16, line 161: its not part of CustomerIOPlugin anymore

PR #201, opened 2025-02-12
chore: Public release builds

mahmoud-elmorabea:
This PR creates a new workflow that publishes a test app that uses the latest published CIO Flutter SDK version. This workflow can be launched manually.

The published test apps on Android and iOS have the same version as the SDK version they were built with to make it easy to find the desired version for the tester.

There can be multiple test app versions on the Firebase Distribution `public` channel with the same version in case we trigger this workflow for the same SDK version but with new changes to the test apps.


apps/amiapp_flutter/pubspec.yaml
```diff
@@ -47,6 +47,8 @@ dependencies:
   firebase_core: ^3.3.0
   firebase_messaging: ^15.0.4
   flutter_local_notifications: ^17.1.2
+  dotenv: ^4.2.0
```
Shahroz16, line 50: we are already using `flutter_dotenv`, whats difference in these?

Shahroz16 (approved):

apps/amiapp_flutter/scripts/update_sample_app_sdk_version.dart
```diff
@@ -0,0 +1,32 @@
+// ignore_for_file: avoid_print
+import 'dart:io';
+import 'package:dotenv/dotenv.dart';
+import 'package:path/path.dart' as path;
+
+void main() async {
+  final scriptDir = File(Platform.script.toFilePath()).parent;
```
mrehan27, line 7: The script looks good. It would be helpful to add some inline comments explaining how it works to make future maintenance easier.

apps/fastlane/Fastfile
```diff
@@ -33,7 +33,7 @@ lane :update_flutter_sdk_version do |options|
 end
 
 # Lane to update Flutter app version
-lane :update_flutter_app_version do |options|
+lane :update_flutter_android_app_version do |options|
```
mrehan27, line 36: We could use `options` and reuse the same for Android and iOS I believe? 🤔

apps/amiapp_flutter/scripts/update_sample_app_sdk_version.dart
mahmoud-elmorabea: Lemme know if there is anything specific that's not clear and I can comment on it in a follow up maybe?

apps/fastlane/Fastfile
mahmoud-elmorabea: Not sure if that's going to work, because the iOS lanes fail when included in a workflow that runs on Linux

PR #218, opened 2025-04-15
fix: Fix in-app module initialization to use region US by default

mahmoud-elmorabea:
Closes: [MBL-1052](https://linear.app/customerio/issue/MBL-1052/flutter-sdk-region-is-not-set-correctly-by-default)

## Problem
In case of a customer omitting the `region` property from their SDK config, initializing in-app module failed.

The flutter bridge to iOS native SDK that is initializing In-App module is using an extension `MessagingInAppConfigBuilder.from([String: Any?])`. It was attempting to create `MessagingInAppConfigOptions` but it silently failed when an error is thrown.

## Solution
- I've added some logic to extract `siteId` and `region` from the Flutter config before passing it in the bridge to iOS native calls.
- The logic will print appropriate error logs in case module initialization fails
- The bridge now uses the more robust typed constructor of `MessagingInAppConfigBuilder` to build `MessagingInAppConfigOptions`

## How to test?
In the Flutter sample app remove the [region](https://github.com/customerio/customerio-flutter/blob/39bd65ae0bb0f7de3b8aa09bfbca96734e0e1f52/apps/amiapp_flutter/lib/src/customer_io.dart#L67) from the Flutter config and try running the app

The app should start normally and in-app module should be initialized successfully and you are able to receive in-app messages.

### DIFF
--- a/ios/Classes/MessagingInApp/CustomerIOInAppMessaging.swift
+++ b/ios/Classes/MessagingInApp/CustomerIOInAppMessaging.swift
@@ -40,10 +40,23 @@ public class CustomerIOInAppMessaging: NSObject, FlutterPlugin {
     }
 
     func configureModule(params: [String: AnyHashable]) {
-        if let inAppConfig = try? MessagingInAppConfigBuilder.build(from: params) {
-            MessagingInApp.initialize(withConfig: inAppConfig)
-            MessagingInApp.shared.setEventListener(CustomerIOInAppEventListener(invokeDartMethod: invokeDartMethod))
+        guard let rawInAppConfig = params["inApp"] else {
+            return
         }
+        guard let inAppConfig = rawInAppConfig as? [String: Any] else {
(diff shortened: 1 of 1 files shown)


mrehan27 (changes requested):

ios/Classes/MessagingInApp/CustomerIOInAppMessaging.swift
```diff
@@ -40,10 +40,23 @@ public class CustomerIOInAppMessaging: NSObject, FlutterPlugin {
     }
 
     func configureModule(params: [String: AnyHashable]) {
```
Shahroz16, line 42: lets add test for this method and I think that would suffice all concerns, of parsing

Shahroz16: We do have tests in flutter repo but they don't go till binding (swift) level, but I don't expect it to be hard either (I can be wrong). I think we have channels setup for test, we just need to see how can we listen to those test channels at native level.

PR #219, opened 2025-04-16
chore: Trigger sample apps public builds after deployment is successful

mahmoud-elmorabea:
Closes: [MBL-1053](https://linear.app/customerio/issue/MBL-1053/auto-create-testbed-releases-for-every-sdk-release)

This PR triggers a sample apps public release build after the SDK has been deployed successfully so that internal stakeholders always have up to date versions to test with.

### Note
I am not entirely sure if the package will be immediately available after publishing before it can be used to build the sample app or we need a delay. Starting out with immediate trigger and we can revise later if the build fails.

### DIFF
--- a/.github/workflows/build-sample-app-for-sdk-release.yml
+++ b/.github/workflows/build-sample-app-for-sdk-release.yml
@@ -2,6 +2,7 @@ name: Publish test apps for SDK release
 
 on:
   workflow_dispatch:
+  workflow_call:
 
 jobs:
   build-sample-apps:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-sample-app-for-sdk-release.yml
```diff
@@ -2,6 +2,7 @@ name: Publish test apps for SDK release
 
 on:
   workflow_dispatch:
+  workflow_call:
```
mahmoud-elmorabea, line 5: This is required for [reusable workflows](https://docs.github.com/en/actions/sharing-automations/reusing-workflows#creating-a-reusable-workflow).

Previously, this workflow was only manually triggered which only needs `workflow_dispatch` but to be able to call it from another workflow we need it to respond to `workflow_call` as well

Shahroz16 (approved):

PR #220, opened 2025-04-17
chore: Log error if swift bridge fails to create MessagingInAppConfigBuilder

mahmoud-elmorabea:
Closes: [MBL-1052](https://linear.app/customerio/issue/MBL-1052/flutter-sdk-region-is-not-set-correctly-by-default)

## Problem
In case of a customer omitting the `region` property from their SDK config, initializing in-app module failed.

The flutter bridge to iOS native SDK that is initializing In-App module is using an extension `MessagingInAppConfigBuilder.from([String: Any?])`. It was attempting to create `MessagingInAppConfigOptions` but it silently failed when an error is thrown.

## Solution
The Flutter bridge shouldn't silently fail and should log an error that initializing in-app module has failed


Shahroz16 (commented):
looks good, lets wait for Rehans PRs on iOS SDK, so we can bump the version number in the same PR to avoid multiple releases

mahmoud-elmorabea:
> looks good, lets wait for Rehans PRs on iOS SDK, so we can bump the version number in the same PR to avoid multiple releases

@Shahroz16 changed PR to `chore`, we can do the version bump separately

Shahroz16 (approved):

PR #232, opened 2025-06-21
chore:  use shared wrapper classes for platform events

Shahroz16:
Refactor Flutter inline in-app messaging to use shared wrapper classes from the native Customer.io SDK, eliminating code duplication between React Native and Flutter platforms.

  **Key Changes:**
  - Replace custom `BaseInlineInAppMessageView` usage with `WrapperInlineView<FlutterInAppPlatformDelegate>`
  - Replace custom `AndroidInAppPlatformDelegate` usage with `WrapperPlatformDelegate`
  - Consolidate Flutter-specific logic to only event dispatch implementation
  - All animation, state management, and lifecycle logic now shared across platforms

  **Files Changed:**
  - `android/build.gradle` - Currently 
  - `FlutterInlineInAppMessageView.kt` - Now extends `WrapperInlineView` with shared lifecycle management
  - `FlutterInAppPlatformDelegate.kt` - Now extends `WrapperPlatformDelegate` with only Flutter-specific event dispatch

  ## Benefits

  - **Reduced Code Duplication**: Shared implementation between React Native and Flutter
  - **Improved Maintainability**: Single source of truth for animation and state logic
  - **Consistent Behavior**: Identical behavior across all wrapper platforms
  - **Simplified Codebase**: Platform-specific code reduced to essential event dispatch only

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -59,7 +59,7 @@ android {
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
     // Customer.io SDK
-    def cioVersion = "4.6.1"
+    def cioVersion = "local"
     implementation "io.customer.android:datapipelines:$cioVersion"
     implementation "io.customer.android:messaging-push-fcm:$cioVersion"
     implementation "io.customer.android:messaging-in-app:$cioVersion"
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -18,6 +18,7 @@ rootProject.allprojects {
     repositories {
         google()
         mavenCentral()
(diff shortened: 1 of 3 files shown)


android/build.gradle
```diff
@@ -59,7 +60,7 @@ android {
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
     // Customer.io SDK
-    def cioVersion = "4.6.1"
+    def cioVersion = "consoladated-platform-SNAPSHOT" // Replace with the actual version after native release
```
mahmoud-elmorabea, line 63: Is this temporary while testing? We should probably create a ticket or a release checklist to capture these things to be reverted before release if not already have one

PR #365, opened 2026-07-18
feat(inbox): Flutter Visual Notification Inbox UI + InboxEventListener bridge

mahmoud-elmorabea:
Exposes the Visual Notification Inbox UI (overlay, bell, view) to Flutter and bridges the global `InboxEventListener` (host action + observational events) across iOS and Android.

> Note: CI is expected to be red until the native inbox modules are published — native pins are intentionally still local-dev/unreleased on this draft.

---

> [!NOTE]
> **Medium Risk**
> Large cross-platform surface (Compose/SwiftUI platform views, single process-wide inbox listener, host-owned navigation) plus temporary snapshot/branch native dependencies; behavior is well-documented but integration mistakes can break inbox taps or multi-engine setups.
> 
> **Overview**
> Adds **Visual Notification Inbox** to the Flutter plugin: native bell and message-list UI via platform views, plus a bridged global **`InboxEventListener`** so Dart can receive inbox lifecycle events and own action navigation.
> 
> **Dart API:** New `NotificationInboxBell` and `NotificationInboxView` widgets (exported from `customer_io_widgets`), `InboxEventListener`, and `CustomerIO.inAppMessaging.setInboxEventListener`. While a listener is registered, native always treats actions as host-handled (async channel cannot return a bool), so apps must handle URLs/deeplinks themselves. Method-channel registration handles init failures and races between early/late registration.
> 
> **Android:** Enables Compose, depends on unreleased `messaging-inbox` snapshot, registers bell/list platform views with lifecycle/recomposer workarounds for `FlutterActivity`, and forwards inbox events with multi-engine-safe forwarder ownership on engine detach.
> 
> **iOS:** SwiftUI-hosted platform views (`NotificationInboxOverlay` / list), bell tap observation, hosting-controller hierarchy fix for sheets, parallel inbox event bridge with weak captures and `INBOX_NOT_AVAILABLE` guard.
> 
> **Tooling / samples:** Podspec and SPM pin unreleased native inbox branches; CocoaPods sample Podfile pulls branch pods; SPM sample adds an **Inbox UI Components** demo screen with `url_launcher` for actions. Public API surface updated in `customerio-flutter.api`; unit tests cover listener register/unregister and failure paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 221ed4cf3b08027f265b6ec8a70fc3bf3789dc82. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lib/messaging_in_app/method_channel.dart
```diff
@@ -37,6 +44,61 @@ class CustomerIOMessagingInAppMethodChannel
...
+  @override
+  void setInboxEventListener(InboxEventListener? listener) {
+    // Always tear down any existing subscription first.
+    _inboxEventSubscription?.cancel();
+    _inboxEventSubscription = null;
+
+    if (listener == null) {
+      methodChannel
+          .invokeNativeMethodVoid(NativeMethods.unregisterInboxEventListener);
+      return;
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Valid — fixed in 6863c98, same treatment as the register path:

```dart
methodChannel
    .invokeMethod<void>(NativeMethods.unregisterInboxEventListener)
    .catchError((Object error) {
  if (kDebugMode) {
    print("Failed to unregister inbox event listener natively: $error");
  }
});
```

Worth noting the fix is observability only, not recovery: the caller asked for no listener, so there is nothing to roll back to — keeping the Dart subscription alive would dispatch to a listener the app has already discarded. What the raw channel buys is that the desync stops being invisible.

The related teardown case you raised in the other thread is the one with real consequences, and that is fixed properly there.

android/src/main/kotlin/io/customer/customer_io/messaginginapp/CustomerIOInAppMessaging.kt
```diff
@@ -99,6 +117,8 @@ internal class CustomerIOInAppMessaging(
             "markInboxMessageUnopened" -> call.nativeMapArgs(result, ::markInboxMessageUnopened)
             "markInboxMessageDeleted" -> call.nativeMapArgs(result, ::markInboxMessageDeleted)
             "trackInboxMessageClicked" -> call.nativeMapArgs(result, ::trackInboxMessageClicked)
+            "registerInboxEventListener" -> registerInboxEventListener(result)
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Valid on both platforms — fixed in 6863c98. Good catch; this was the more serious of the two.

**Android** — `onDetachedFromEngine` cleared only the change listener:

```kotlin
override fun onDetachedFromEngine() {
    clearInboxChangeListener()
    clearInboxEventListener()
    super.onDetachedFromEngine()
}
```

`clearInboxEventListener()` is guarded by an `isInboxEventListenerRegistered` flag (mirroring the existing `isInboxChangeListenerSetup`) so an engine that never used the inbox cannot clear another engine's forwarder.

**iOS** — the equivalent teardown point is `deinit`, but it turned out to be unreachable, which is worth recording since it made the first version of this fix dead code. `DefaultNotificationInbox` holds `inboxEventListener` **strongly**, and `CustomerIOInboxEventListener(invokeDartMethod: invokeDartMethod)` passes a bound method that strongly captures the plugin — so registering a listener retained the plugin for the life of the process and `deinit` never ran. Both SDK listeners now capture `[weak self]`:

```swift
MessagingInApp.shared.setInboxEventListener(
    CustomerIOInboxEventListener { [weak self] method, args in
        self?.invokeDartMethod(method, args)
    }
)
```

The in-app event listener in `configureModule` had the identical shape and had to change too, otherwise it alone kept the plugin alive and `deinit` still never fired. `deinit` then clears the inbox forwarder, guarded by the same "did we install one" flag. `GistDelegate.eventListener` is strong as well, so this also stops the plugin outliving its engine in the in-app path.

Verified: `:customer_io:compileDebugKotlin` BUILD SUCCESSFUL; the iOS `customer_io` Swift target compiles clean for arm64 and x86_64; `flutter analyze` clean; 90 Dart tests pass.

lib/messaging_in_app/method_channel.dart
```diff
@@ -37,6 +44,72 @@ class CustomerIOMessagingInAppMethodChannel
...
+    // arrive. On failure the Dart side is undone so the two stay in agreement.
+    methodChannel
+        .invokeMethod<void>(NativeMethods.registerInboxEventListener)
+        .catchError((Object error) {
+      if (kDebugMode) {
+        print('Failed to register inbox event listener natively: $error');
+      }
+      _inboxEventSubscription?.cancel();
+      _inboxEventSubscription = null;
+    });
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Correct, and my bug — thanks. Fixed in e984480.

The handler cleared the field rather than its own subscription:

```dart
final subscription = _inboxEventStreamController.stream.listen(...);
_inboxEventSubscription = subscription;

methodChannel
    .invokeMethod<void>(NativeMethods.registerInboxEventListener)
    .catchError((Object error) {
  if (kDebugMode) { print("Failed to register inbox event listener natively: $error"); }
  if (!identical(_inboxEventSubscription, subscription)) {
    return;
  }
  subscription.cancel();
  _inboxEventSubscription = null;
});
```

Nothing to do when superseded: the later call already cancelled this subscription at the top of `setInboxEventListener`.

Your ordering premise is exactly why this is reachable rather than theoretical — registration only fails while the SDK is uninitialized, so failures cluster before initialize and successes after, which is precisely the "earlier fails, later succeeds" interleaving.

Worth noting the mirror case is not reachable: for a later attempt to fail after an earlier one succeeded, `CustomerIO.instance()` would have to start throwing again, and there is no de-initialize path. So a failed attempt can never leave a forwarder installed behind it, and unwinding the Dart side is sufficient.

Covered by a test that drives the interleaving (first register throws, second succeeds); with the identity check removed it fails on `Actual: []`.

ios/customer_io/Sources/customer_io/MessagingInbox/NotificationInboxPlatformViews.swift
```diff
@@ -0,0 +1,198 @@
...
+
+/// Hosts a SwiftUI view inside a Flutter `UIView` container.
+///
+/// The inbox SwiftUI components require iOS 15 (Jist floor); on older versions an empty container is
+/// returned so the host app does not crash.
+private func makeHostedContainer<Content: View>(
+    frame: CGRect,
+    @ViewBuilder content: () -> Content
+) -> UIView {
+    guard #available(iOS 15.0, *) else {
```
Shahroz16: Could we remove the iOS 15 gate for the standalone list? The current native SDK declares iOS 13 support and NotificationInboxView is explicitly available from iOS 13; only NotificationInboxOverlay used by the bell requires iOS 16. Because the Flutter CocoaPods podspec also supports iOS 13, this helper currently returns an empty view on iOS 13–14, so the list silently renders blank. I think this should be fixed before merging.

ios/customer_io/Sources/customer_io/MessagingInApp/CustomerIOInAppMessaging.swift
```diff
@@ -65,6 +85,24 @@ public class CustomerIOInAppMessaging: NSObject, FlutterPlugin {
         case "trackInboxMessageClicked":
             trackInboxMessageClicked(call: call, result: result)
 
+        case "registerInboxEventListener":
+            // `[weak self]` is load-bearing: the SDK holds the listener strongly, so passing
+            // `invokeDartMethod` directly would have the listener retain this plugin for the life of
+            // the process. `deinit` would then never run, and neither would the teardown that
+            // restores the SDK's default action handling.
+            MessagingInApp.shared.setInboxEventListener(
```
Shahroz16: Could iOS return INBOX_NOT_AVAILABLE when MessagingInApp.shared.implementation is nil, as Android does? The native setter uses optional chaining, so registration before CustomerIO.initialize() currently succeeds from Dart but does nothing on iOS. The Dart subscription then remains active and there is no automatic retry, so the app believes the listener is registered but never receives events. This startup ordering is also the case the new rollback test covers, so I think this should be fixed before merging.

lib/messaging_in_app/inbox_event_listener.dart
```diff
@@ -0,0 +1,55 @@
+import 'inbox_message.dart';
+
+/// Listener the host app registers to be notified when events occur on Visual
+/// Notification Inbox messages.
+///
+/// Mirrors the native `InboxEventListener` (iOS/Android). Register it via
+/// `CustomerIO.instance.inAppMessaging.setInboxEventListener(listener)` and
```
Shahroz16: Could we change this example to CustomerIO.inAppMessaging.setInboxEventListener(listener)? inAppMessaging is a static getter, so accessing it through CustomerIO.instance does not compile. This is documentation-only and not a blocker, but the public example would fail when copied.

DEVELOPING_LOCALLY.md
```diff
@@ -0,0 +1,156 @@
+# Developing locally — Visual Notification Inbox UI spike
```
Shahroz16: Could we update or remove this file before merging? It still describes the deleted NotificationInboxOverlay callback, personal paths, mavenLocal, older branch names, and an SPM build that is no longer blocked. Several current WIP comments point contributors here, so the person doing the release-pin cleanup would be sent down the wrong path. This is not a runtime blocker, but the new guide should match the code we are shipping.

ios/customer_io/Sources/customer_io/MessagingInbox/NotificationInboxPlatformViews.swift
mahmoud-elmorabea: Fixed in dc90dbb — you were right, and the comment justifying the gate was wrong too: Jist's floor is iOS 13, not 15, so there was never a 15 requirement to honour.

`makeHostedContainer` now carries no availability floor of its own. Each caller gates on what its own content needs: the list is unguarded (`NotificationInboxView` is `@available(iOS 13.0, *)`), and only the bell keeps `if #available(iOS 16.0, *)` for `NotificationInboxOverlay`.

ios/customer_io/Sources/customer_io/MessagingInApp/CustomerIOInAppMessaging.swift
mahmoud-elmorabea: Fixed in dc90dbb. `registerInboxEventListener` is now its own method that guards on `MessagingInApp.shared.hasBeenInitialized` and returns the same `INBOX_NOT_AVAILABLE` code and message Android sends.

Worth noting the Dart side already expected this: the existing rollback tests simulate exactly this `PlatformException`, so iOS was the only side not producing it.

lib/messaging_in_app/inbox_event_listener.dart
mahmoud-elmorabea: Fixed in dc90dbb — now `CustomerIO.inAppMessaging.setInboxEventListener(listener)`, with a parenthetical that `inAppMessaging` is a static getter, since that is the part that bites when the snippet is copied.

DEVELOPING_LOCALLY.md
mahmoud-elmorabea: Removed in 498d20c rather than rewritten — it was a spike log, and every part of it had gone stale.

The single pointer to it (in `Package.swift`) is gone too. Nothing was lost: the revert recipe it referenced is already spelled out inline in that same comment block.

ios/customer_io/Package.swift
```diff
@@ -95,8 +97,21 @@ let package = Package(
...
+        // not resolve a package that one dependency requires by BRANCH while another requires it by
+        // VERSION, and fcm's `main` pins customerio-ios with `from: "4.0.0"`. The scaffold branch
+        // `inbox/overlay-inbox-pin` matches the branch requirement instead, collapsing both to one
+        // package identity with no version requirement.
+        //
+        // Unlike the local paths this replaces, these resolve on CI and on any machine.
+        // On native release, revert both to version pins and delete the fcm scaffold branch:
+        //   .package(url: ".../customerio-ios.git", exact: "4.6.1"),   // per main today
+        //   .package(url: ".../customerio-ios-fcm.git", from: "1.0.0")
+        .package(url: "https://github.com/customerio/customerio-ios.git", branch: "feat/overlay-inbox"),
```
Shahroz16, line 113: The remaining iOS CI failures look like release-pin cleanup rather than a problem with this implementation: SwiftPM and the CocoaPods sample still point to the deleted `feat/overlay-inbox` branch, while Android still uses the feature snapshot. I am okay with this PR merging into `feat/overlay-inbox`, but before that feature branch is merged to `main` or released, could we switch these to customerio-ios 4.7.0 and customerio-android 4.20.0 and confirm the consumer builds pass?

Shahroz16 (approved):

PR #368, opened 2026-07-23
feat(live-activities): expose Live Activities through CustomerIO.liveActivities

mahmoud-elmorabea:
Exposes Live Activities (iOS) / Live Notifications (Android) on the Flutter SDK, consuming the native SDK-managed module so registration happens at SDK init.

Config lives under `liveNotifications` — `types` are the reverse-DNS identifiers shared with both native SDKs; unrecognized ones are ignored. `end` optionally takes a final content-state so iOS can render a terminal card.

Depends on customerio-ios#1178.

---

> [!NOTE]
> **Medium Risk**
> Touches push/notification initialization and SDK versioning with prerelease native dependencies; iOS behavior differs from Android for updates on activities not started in-session.
> 
> **Overview**
> Adds **Live Activities (iOS) / Live Notifications (Android)** to the Flutter SDK via `CustomerIO.liveActivities` (`start` / `update` / `end`), `LiveActivitiesConfig` on `CustomerIOConfig` (`liveNotifications`), and typed `LiveActivityPayload` templates plus a `custom` path keyed by `customType`.
> 
> On **Android**, a new `CustomerIOLiveActivities` module maps payloads to the FCM push module, merges `liveNotifications` into push config at init, and supports a pre-init `setLiveNotificationCallback` for custom rendering. On **iOS**, an opt-in `liveactivities` pod/SPM flag registers `LiveActivitiesModule` at SDK init and exposes `CustomerIOLiveActivities.handleWidgetUrl` for tap attribution.
> 
> **README** documents opt-in steps, `NSSupportsLiveActivities`, Widget Extension requirements, and `AppDelegate` URL forwarding. **Sample apps** gain Live Activity widget extensions, rideshare demos, shared `MainApplication` for Android custom notifications, and **REL-1** temporary pins to unreleased native SDK branches/snapshots until Live Activities ship.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f737bfa72d146e0ec6f10f25580f2405326dbe74. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,9 +18,13 @@ You'll find our complete [SDK documentation here](https://customer.io/docs/sdk/f
 
 Enable the activity types you use via `liveNotificationsConfig` in your `CustomerIOConfig`. On iOS, Live Activities are opt-in: add the `liveactivities` pod subspec (CocoaPods) or set `customerio_live_activities_enabled=true` in `android/gradle.properties` (Swift Package Manager), plus a Widget Extension that renders the SDK's built-in templates.
 
-**One manual step is required on iOS.** Forward every opened URL to the SDK from your `AppDelegate`, or taps on a Live Activity are not attributed:
+For an activity of your own, set `customType` to your reverse-DNS identifier and start it with `LiveActivityPayload.custom(data: {...})`. You supply the view: `CIOCustomAttributes` in your iOS Widget Extension, and the `createLiveNotification` callback on Android.
+
+**One manual step is required on iOS.** Forward every opened URL to the SDK from your `AppDelegate`, or taps on a Live Activity are not attributed. `CustomerIOLiveActivities` comes from the plugin, so import it — and note this only compiles once Live Activities are opted in above:
 
 ```swift
+import customer_io
+
 override func application(_ app: UIApplication, open url: URL, options: [UIApplication.OpenURLOptionsKey: Any] = [:]) -> Bool {
   // Reports an `opened` metric and returns the deep link to route to. A non-Customer.io URL comes
   // back unchanged; `nil` means the activity carried no deep link, so there is nothing to open.
--- a/README.md
+++ b/README.md
@@ -18,6 +18,13 @@ You'll find our complete [SDK documentation here](https://customer.io/docs/sdk/f
 
 Enable the activity types you use via `liveNotificationsConfig` in your `CustomerIOConfig`. On iOS, Live Activities are opt-in: add the `liveactivities` pod subspec (CocoaPods) or set `customerio_live_activities_enabled=true` in `android/gradle.properties` (Swift Package Manager), plus a Widget Extension that renders the SDK's built-in templates.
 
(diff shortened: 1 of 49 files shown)


android/src/main/kotlin/io/customer/customer_io/liveactivities/CustomerIOLiveActivities.kt
```diff
@@ -0,0 +1,323 @@
...
+            customActivityType = customType
+            if (customType != null) {
+                builder.enableCustomLiveNotificationTypes(customType)
+            }
+
+            val branding = config.getAs<Map<String, Any>>("branding")
+            if (branding != null) {
+                val accentColor = branding.getAs<String>("accentColorHex")
+                    ?.let { runCatching { Color.parseColor(it) }.getOrNull() }
+                    ?: Color.TRANSPARENT
```
Shahroz16, line 293: `accentColorHex` is optional, so could we preserve the absence of a color here instead of passing `Color.TRANSPARENT` as an explicit override? Android treats that zero value as its default color, so this does not make the UI invisible, but it does bypass an app's configured `default_notification_color` whenever branding only supplies a logo or company name. This feels non-blocking, but keeping the value nullable through the native contract would preserve the normal app-level fallback.

README.md
```diff
@@ -14,6 +14,27 @@ This is the official Customer.io Flutter plugin.
 
 You'll find our complete [SDK documentation here](https://customer.io/docs/sdk/flutter/). 
 
+# Live Activities
+
+Enable the activity types you use via `liveNotificationsConfig` in your `CustomerIOConfig`. On iOS, Live Activities are opt-in: add the `liveactivities` pod subspec (CocoaPods) or set `customerio_live_activities_enabled=true` in `android/gradle.properties` (Swift Package Manager), plus a Widget Extension that renders the SDK's built-in templates.
```
Shahroz16, line 19: Could we add `NSSupportsLiveActivities = true` to the iOS setup here? The sample adds it, but the public instructions only mention the dependency, opt-in flag, widget and URL forwarding. Without the host Info.plist key ActivityKit will not start an activity even though everything compiles, so this is a required setup step.

pubspec.yaml
```diff
@@ -43,5 +43,5 @@ flutter:
         pluginClass: CustomerIOPlugin
       ios:
         pluginClass: CustomerIOPlugin
-        native_sdk_version: 4.5.3
+        native_sdk_version: 4.6.0
```
Shahroz16: Can we update this pin to the native release that will actually contain Live Activities before publishing the plugin? The current iOS aggregate is already at 4.6.1, while 4.6.0 predates the feature and the new Live Activities podspecs on the branch currently conflict with the rest of 4.6.1. The sample's branch override exposes that conflict today; once the native podspecs are aligned, this pin needs to move with them.

Shahroz16 (approved):

PR #369, opened 2026-07-23
feat(sample): Live Activities demo screens

mahmoud-elmorabea:
Demo screens for the two built-in templates and an app-owned custom type, plus AppDelegate routing for the opened metric.

Stacked on #368.

---

> [!NOTE]
> **Low Risk**
> Sample-app and demo wiring only; AppDelegate URL handling is additive and preserves non–Live Activity deep links. Temporary iOS SDK branch pins are intentional for pre-release testing.
> 
> **Overview**
> **Adds Live Activities end-to-end demos** to the CocoaPods and SPM Flutter sample apps: segments and countdown built-in templates plus a custom **rideshare** type, wired through native layers and a new Dart test screen (SPM app).
> 
> On **Flutter (SPM sample)**, SDK init gains `LiveActivitiesConfig` (built-in templates + custom `rideshare` type), a **Live Activities** dashboard entry, routing to `LiveActivitiesScreen`, and start/update/end flows for all three activity kinds via `CustomerIO.liveActivities`.
> 
> On **iOS (both samples)**, a **LiveActivityWidget** extension registers SDK template widgets and app-owned `RideshareLiveActivity` SwiftUI; Runner enables `NSSupportsLiveActivities`, Fastlane lists the widget bundle ID, and **AppDelegate** routes widget taps through `CustomerIOLiveActivities.handleWidgetUrl` before Flutter deep links. CocoaPods adds `liveactivities` plus temporary `feat/live-activities` git pod overrides; SPM pulls the same branch via Swift Package.
> 
> On **Android (both samples)**, manifests point at shared **`MainApplication`**, which registers a **rideshare** `CustomerIOLiveNotificationsCallback` for custom live notifications when the process starts (including cold-start push), with `customerio_live_activities_enabled=true` in Gradle.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 969eff023df73d2d927b9ee6f3069a8e89852f2a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


apps/flutter_sample_cocoapods/ios/LiveActivityWidget/RideshareLiveActivity.swift
```diff
@@ -0,0 +1,49 @@
...
+                    Text(context.state.data["status"] ?? "")
+                }
+            } compactLeading: {
+                Image(systemName: "car.fill")
+            } compactTrailing: {
+                Text("\(context.state.data["etaMinutes"] ?? "")m")
+            } minimal: {
+                Image(systemName: "car.fill")
+            }
+        }
```
Shahroz16, line 55: Could we apply `.cioWidgetUrl(context.state.cioMetadata)` to both the lock-screen view and the `DynamicIsland`, like the bundled templates do? The AppDelegate routing in this PR is never reached for a Rideshare tap without that URL, so every server-pushed custom activity loses its opened metric and customer deep-link.

Shahroz16 (commented):
I confirmed the existing Cursor thread about the SPM sample is practical: the SPM app owns the shared Dart screen but has none of the native iOS wiring and no Android custom callback. I am not duplicating that thread. I found one additional custom-widget gap below. Android custom tap and dismissal behavior is covered by the root SDK comment on Android #734.

apps/flutter_sample_spm/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -439,6 +540,26 @@
 			};
 			name = Profile;
 		};
+		272C8EE873213FBBD7CF6B67 /* Release */ = {
+			isa = XCBuildConfiguration;
+			buildSettings = {
+				CLANG_ENABLE_OBJC_WEAK = NO;
+				CODE_SIGNING_ALLOWED = NO;
```
Shahroz16: This is still blocking the SPM iOS sample build. The current CI run gets through compilation and then fails because the embedded widget is not signed with the same certificate as Runner. Fastlane successfully installs the profile for io.customer.testbed.flutter.spm.LiveActivityWidget, but this setting prevents Xcode from signing the extension with it. Could we let this target use the same signing setup as Runner and the notification extension? Right now the primary sample cannot produce an installable iOS build.

apps/flutter_sample_spm/android/app/src/main/kotlin/io/customer/testbed/flutter/spm/MainActivity.kt
```diff
@@ -8,6 +18,15 @@ import io.flutter.plugin.common.MethodChannel
 class MainActivity : FlutterActivity() {
     private lateinit var permissionHandler: PermissionChannelHandler
 
+    override fun onCreate(savedInstanceState: Bundle?) {
```
Shahroz16: Could we cover the process-cold FCM path before calling this end to end? Right now the custom callback is registered only from `MainActivity`, and the enabled types are applied only when Dart runs `CustomerIO.initialize`. When Android recreates a killed process for `CustomerIOFirebaseMessagingService`, neither of those runs. The native service only sets up the context, so `pushModuleConfig` falls back to its default empty type list/null callback and drops the start, update, or end at the allow-list check. This happens for every Live Notification delivered after normal process death; a dropped end can also leave the previous ongoing notification stranded. The underlying bootstrap needs to be solved in #368, and then this sample should register its custom renderer through that process-start path rather than the Activity.

apps/flutter_sample_spm/android/gradle.properties
```diff
@@ -2,6 +2,7 @@ org.gradle.jvmargs=-Xmx4g
 android.useAndroidX=true
 android.enableJetifier=true
 customerio_location_enabled=true
+customerio_live_activities_enabled=true
```
Shahroz16, line 5: Could we also add `android.permission.POST_PROMOTED_NOTIFICATIONS` to both sample manifests? The native SDK deliberately leaves this permission to the host app. On Android 16+, without it the Segments and Countdown notifications still render, but only as regular ongoing notifications, so this demo never exercises the promoted Live Notification treatment. This one is non-blocking, but it is a one-line sample setup requirement.

Shahroz16 (commented):
Re-reviewed the current head. The earlier widget URL and SPM signing issues are fixed, all four sample builds are green, and the CocoaPods app does share this Dart screen through its symlink, so that Cursor comment is not valid. I found one real Android cold-process gap in the stacked initialization path.

Shahroz16 (approved):

PR #372, opened 2026-07-29
feat(sample): Visual Inbox demo screen in the Flutter sample app

mahmoud-elmorabea:
Stacked on #365 (SDK). Adds only the demo UI — no SDK changes.

A Visual Inbox screen in the SPM sample exercising both inbox widgets — the branded bell, whose tap opens the SDK's own panel, and the embedded message list — plus an on-screen readout of the `InboxEventListener` callbacks so the listener bridge can be checked by hand. Tapping a message action navigates via `url_launcher`: registering a listener makes the host own action navigation, so the demo has to do it.

Build wiring the samples need to resolve the unreleased native inbox — the cocoapods Podfile branch pins — deliberately lives in the SDK PR so that one stands on its own in CI.

Merge #365 first.

---

> [!NOTE]
> **Low Risk**
> Sample-app-only navigation and demo UI; no production SDK or auth changes in this diff.
> 
> **Overview**
> Adds an **Inbox UI Components** flow to the Flutter SPM sample so Visual Notification Inbox PlatformViews and the `InboxEventListener` bridge can be exercised manually.
> 
> A new `InboxUiScreen` (route `inbox-ui`, dashboard entry) uses tabs for **`NotificationInboxBell`** (observational `onTap` plus on-screen status) and an embedded **`NotificationInboxView`**. While the screen is open it registers a demo **`InboxEventListener`** that surfaces callback summaries and opens action URLs via **`url_launcher`** (listener registration makes the host own navigation). The listener is cleared on dispose.
> 
> Also bumps the path **`customer_io`** lockfile version in the cocoapods sample and adds **`url_launcher`** to the SPM sample `pubspec.yaml`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 314b569857264aa7432f37f53fcc2636b451566b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/flutter_sample_spm/lib/src/screens/inbox_ui.dart
+++ b/apps/flutter_sample_spm/lib/src/screens/inbox_ui.dart
@@ -1,0 +1,215 @@
+import 'dart:developer';
+
+import 'package:customer_io/customer_io.dart';
+import 'package:customer_io/customer_io_widgets.dart';
+import 'package:customer_io/messaging_in_app.dart';
+import 'package:flutter/material.dart';
+
+import '../components/container.dart';
+
+/// Demonstrates the three native Visual Notification Inbox UI components exposed
+/// via PlatformView:
+///   1. NotificationInboxOverlay (drop-in bell + slide-out panel)
+///   2. NotificationInboxBell    (bell only; host opens its own UI)
+///   3. NotificationInboxView    (the Jist-rendered message list)
+///
+/// Message actions are handled by the global InboxEventListener (bridged
+/// separately); these widgets only render UI and surface widget-level callbacks.
+class InboxUiScreen extends StatefulWidget {
+  const InboxUiScreen({super.key});
+
+  @override
+  State<InboxUiScreen> createState() => _InboxUiScreenState();
+}
+
+class _InboxUiScreenState extends State<InboxUiScreen>
+    with SingleTickerProviderStateMixin {
+  late TabController _tabController;
+
+  /// Last inbox event surfaced by the global [InboxEventListener], shown on-screen.
+  String _lastInboxEvent = 'No inbox events yet';
+
+  @override
+  void initState() {
+    super.initState();
+    _tabController = TabController(length: 3, vsync: this);
+
+    // Register a global inbox event listener. While registered, the Flutter host
+    // owns inbox action navigation (the SDK suppresses its default handling).
+    CustomerIO.inAppMessaging.setInboxEventListener(
+      _DemoInboxEventListener(
+        onEvent: (summary) {
+          log('[InboxEventListener] $summary');
+          if (mounted) {
+            setState(() => _lastInboxEvent = summary);
+          }
+        },
+      ),
+    );
+  }
+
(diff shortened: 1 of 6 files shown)


apps/flutter_sample_spm/lib/src/screens/inbox_ui.dart
```diff
@@ -0,0 +1,173 @@
...
+/// the host owns inbox action navigation (the SDK does not open urls itself).
+class _DemoInboxEventListener implements InboxEventListener {
+  _DemoInboxEventListener({required this.onEvent});
+
+  final void Function(String summary) onEvent;
+
+  @override
+  void messageActionTaken(
+      InboxMessage message, String actionName, String actionValue) {
+    onEvent(
```
Shahroz16, line 196: Could the sample either handle actionValue here or avoid registering the listener? Both native bridges report an action as handled while a Flutter InboxEventListener is installed, which suppresses the SDK’s own URL and deep-link handling. Since this callback only logs the action, tapping an inbox action currently does nothing. I think the sample should demonstrate the required navigation handling before merging.

mahmoud-elmorabea: Fixed in 314b569 — you were right on both the mechanism and the consequence.

The demo now navigates on `messageActionTaken`, mirroring what the React Native sample already does with `Linking.openURL`: the action value is handed to the OS, so this app's own `flutter-spm` scheme round-trips back through its deep-link handling and anything else opens externally. Failures surface in the on-screen readout instead of passing silently.

That needed `url_launcher`, since Flutter has no built-in `Linking` equivalent. No `canLaunchUrl` precheck — on iOS it returns false for any scheme missing from `LSApplicationQueriesSchemes` even when launching would succeed.

Also corrected the PR description, which still described a three-widget screen from before the overlay was removed.

apps/flutter_sample_spm/pubspec.yaml
```diff
@@ -40,6 +40,10 @@ dependencies:
     path: ../../
   email_validator: ^3.0.0
   go_router: ^14.2.3
+  # Inbox action navigation: registering an InboxEventListener makes this app own it, so the demo
+  # hands the action url to the OS the way the SDK would. Flutter has no built-in equivalent of
+  # React Native's Linking.
+  url_launcher: ^6.3.0
```
Shahroz16, line 46: The feature behavior here looks good. The remaining failures are integration CI issues: the shared sample source imports `url_launcher`, but the CocoaPods sample does not declare it; the SPM Android sample resolved a `url_launcher_android` version that its current build toolchain cannot process; and the iOS jobs inherit the deleted native branch pin from #365. I do not see these as blocking this sample PR from merging into the feature stack, but could we correct the sample dependencies and lockfiles, along with the inherited native pins, before the final feature branch is merged or released so both sample integrations are validated?

Shahroz16 (approved):

PR #378, opened 2026-08-04
ci: move iOS sample-app build to macos-26 and fix the ignored Xcode pin

mahmoud-elmorabea:
> ✅ **Unblocked.** [mobile-ci-tools#13](https://github.com/customerio/mobile-ci-tools/pull/13) is merged and the `setup-ios` ref here now points at `@main`, so nothing depends on the transitional `macos-26` branch.

[MBL-2224: Migrate iOS CI to `macos-26` + Xcode 26.6, and stop downloading simulator runtimes we already have](https://linear.app/customerio/issue/MBL-2224/migrate-ios-ci-to-macos-26-xcode-266-and-stop-downloading-simulator)

---

This repo carries the fleet's largest share of a fleet-wide waste: roughly 212 minutes of macOS runner time a week downloading a simulator runtime that is not on the image and that these jobs never use, since they build device archives.

It also pinned an Xcode version that never took effect, because the value was defined in the calling workflow and does not cross a reusable-workflow boundary. The pin silently resolved to "whatever is latest" instead, which is what caused the download in the first place.

Moves the macOS runner from 15 to 26, and makes the Xcode version an explicit, actually-applied pin.

### DIFF
--- a/.github/workflows/build-sample-apps.yml
+++ b/.github/workflows/build-sample-apps.yml
@@ -9,9 +9,6 @@ concurrency: # cancel previous workflow run if one exists.
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
-env:
-  XCODE_VERSION: "26.2"
-
 jobs:
   update-pr-comment:
     if: ${{ github.event_name == 'pull_request' }}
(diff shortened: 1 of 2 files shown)


Shahroz16 (approved):
Think the plan is to when https://github.com/customerio/mobile-ci-tools/pull/13 merges, we replace all three temporary setup-ios@macos-26 references with setup-ios@main and rerun the affected macOS jobs before merging.?

PR #384, opened 2026-08-10
ci: detect native SDK versions correctly in auto-update workflow

mrehan27:
## Problem

The daily auto-update workflow has been opening PRs that change no versions — #383 is the clearest case: a real commit whose entire diff is two regenerated `pubspec.lock` files.

[#371](https://github.com/customerio/customerio-flutter/pull/371) changed the Android version line to support local override builds:

```groovy
def cioVersion = rootProject.findProperty("cioVersion") ?: "4.20.1"
```

Both Android patterns in the workflow assumed the bare `def cioVersion = "<version>"` form:

1. **Extraction** — `sed 's/.*def cioVersion = "\(.*\)"/\1/'` stopped matching and passed the line through unchanged, so `current_android` became the whole line. It never equals `4.20.1`, so `android_needs_update` was `true` every single day.
2. **Update** — same assumption, so the replacement `sed` bumped nothing.

`git add .` then swept up unrelated lockfile churn, giving the PR a commit and hiding the fact that nothing was updated. From [run 31305513311](https://github.com/customerio/customerio-flutter/actions/runs/31305513311):

```
Android:     def cioVersion = rootProject.findProperty("cioVersion") ?: "4.20.1" → 4.20.1 (needs update: true)
```

Two further bugs surfaced from the same PR:

- The PR body was garbled. `--body "${{ ... }}"` let the body's own quote characters break shell word-splitting, which is why quotes and indentation vanished from #383's description.
- Release-notes links pointed at `tag/v4.20.1`, but both native repos tag without a `v` prefix — a 404.

## Changes

All in `.github/workflows/auto-update-native-sdks.yml`.

**The bug**
- Extraction takes the version literal off the matched line instead of assuming its shape, and **fails the job** when it can't parse one rather than comparing a bogus value.
- The Android `sed` replaces the last quoted literal on the line, so both the bare and `findProperty` forms work.
- Every file edit is verified afterwards, so a `sed` that silently stops matching fails loudly instead of producing an empty PR.

**Empty PRs can't happen again**
- `git add` is limited to the five version files, and the step aborts when nothing is staged. This alone would have blocked #383.

**PR content**
- The body is built in a file and passed via `--body-file`; title/body/reason go through `env:` instead of inline `${{ }}`.
- Release links use the tag that actually exists.
- The commit message and body list only the platforms that actually changed — no more `iOS: 4.7.2 → 4.7.2`.
- Dropped "iOS compilation verified" / "Android compilation verified" from the body: that job only runs `flutter analyze` and `flutter test`, so those lines were never true.

**Duplicate detection**
- Previously it skipped only when an open PR had the exact generated title, so an update PR opened by hand went unnoticed and the bot opened its own. It now also scans open PRs' diffs for the specific version bumps, catching any author.

## Verification

The workflow's shell steps were replayed locally against a scratch worktree, with `${{ }}` expressions resolved and real `gh` API calls:

- Today's versions (iOS 4.7.2 / Android 4.20.1) → no update proposed, no PR. Replaying the *old* workflow the same way reproduced the production bug exactly, matching the run log above.
- Downgraded baseline → all five version files updated, correct body generated, exactly 5 files committed with simulated lockfile churn excluded.
- Unparseable version → job fails. `sed` that can't match → job fails.
- Duplicate detection: found #381 by diff under a deliberately non-matching title, and correctly *declined* to match it when both platforms were needed, since #381 only bumps iOS.

`actionlint` + `shellcheck` clean — the 2 pre-existing `SC2309` warnings are gone and no new findings. `dart analyze lib test` clean, 105 tests pass.

`scripts/check_api_changes.sh` reports a diff locally, but it is purely a reordering of two same-named `InAppMessage` blocks — sorted, the files are identical, so the API surface is unchanged. This change touches only workflow YAML and cannot affect Dart extraction; `customerio-flutter.api` is left untouched on this branch.

## Follow-up

#377, #380, #381 and #383 are all products of this bug and should be closed (#377/#383 are lockfile-only; #380/#381 bump iOS but never Android). Not touched here.

---

> [!NOTE]
> **Low Risk**
> Changes are confined to CI workflow shell/YAML; no runtime app code or dependency resolution paths are modified.
> 
> **Overview**
> Fixes the daily **auto-update native SDKs** workflow so it stops opening bogus PRs when Android uses `findProperty("cioVersion") ?: "<version>"` and when `sed` edits fail silently.
> 
> **Version detection and edits** now pull semver literals from matched lines (iOS and Android), **fail the job** on parse errors, use Android `sed` that works for both version line shapes, post-edit **grep verification** on all five pins, and relax the SPM `project.pbxproj` replace to handle suffixed versions.
> 
> **Commits and PRs** stage only the five version files, **abort** if nothing changed, build the PR body in a temp file with **`--body-file`** and `env:` for shell safety, resolve **release tag** URLs without assuming a `v` prefix, and list only platforms that actually bump in commit/PR text (with honest testing bullets).
> 
> **Duplicate detection** still matches identical titles but also scans open PR **diffs** for the target iOS/Android version bumps so hand-opened update PRs are not duplicated.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c0f125f485ed091475ca97e1ea12786479ad3368. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/auto-update-native-sdks.yml
+++ b/.github/workflows/auto-update-native-sdks.yml
@@ -75,14 +75,25 @@ jobs:
       - name: Get current SDK versions
         id: current_versions
         run: |
-          # Get current iOS version from pubspec.yaml
-          current_ios=$(grep 'native_sdk_version:' pubspec.yaml | sed 's/.*native_sdk_version: *//')
+          # Takes the version literal off the matched line instead of assuming its shape, so
+          # wrappers like `findProperty("cioVersion") ?: "<version>"` still parse. Failing to
+          # parse must fail the job: falling back to the raw line reports a bogus update.
+          extract_version() {
+            local label=$1 file=$2 pattern=$3 version
+            version=$(grep -m1 -- "$pattern" "$file" | grep -oE '[0-9]+\.[0-9]+\.[0-9]+[0-9A-Za-z.+-]*' | tail -1)
+            if [[ ! "$version" =~ ^[0-9]+\.[0-9]+\.[0-9]+ ]]; then
+              echo "❌ Could not parse $label version from $file (matched '$pattern', got '$version')" >&2
+              exit 1
+            fi
+            echo "$version"
+          }
+
+          current_ios=$(extract_version "iOS" pubspec.yaml 'native_sdk_version:') || exit 1
           echo "current_ios=$current_ios" >> $GITHUB_OUTPUT
-          
-          # Get current Android version from build.gradle
-          current_android=$(grep 'def cioVersion = ' android/build.gradle | sed 's/.*def cioVersion = "\(.*\)"/\1/')
+
+          current_android=$(extract_version "Android" android/build.gradle 'def cioVersion') || exit 1
           echo "current_android=$current_android" >> $GITHUB_OUTPUT
-          
+
           echo "📋 Current versions:"
           echo "iOS: $current_ios"
           echo "Android: $current_android"
@@ -99,7 +110,7 @@ jobs:
             local repo=$1
             gh api "repos/$repo/releases/latest" --jq '.tag_name' 2>/dev/null | sed 's/^v//' || echo ""
           }
-          
+
           CURRENT_IOS="${{ steps.current_versions.outputs.current_ios }}"
           CURRENT_ANDROID="${{ steps.current_versions.outputs.current_android }}"
           INPUT_IOS="${{ github.event.inputs.ios_version }}"
@@ -116,7 +127,7 @@ jobs:
           if [[ -n "$LATEST_IOS" && "$LATEST_IOS" != "$CURRENT_IOS" ]]; then
             IOS_NEEDS_UPDATE="true"
           fi
-          
+
           if [[ -n "$LATEST_ANDROID" && "$LATEST_ANDROID" != "$CURRENT_ANDROID" ]]; then
             ANDROID_NEEDS_UPDATE="true"
           fi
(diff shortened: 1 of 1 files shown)


Shahroz16 (approved):
Looks good to merge.

Non-blocking FYI: the workflow accepts suffixed semantic versions, but the Xcode project update regex only matches plain `X.Y.Z`, and the `grep -E` patterns escape only dots. A manual beta-to-beta or build-metadata update can therefore fail verification. This does not affect the current stable-release path, but it would be worth aligning the accepted version grammar and escaping in a follow-up.

mrehan27:
@Shahroz16 thanks — both addressed in c0f125f:

- `project.pbxproj`: the sed now matches `version = [^;]+;`, so a suffixed current value gets replaced. The search side previously required a plain `X.Y.Z` and silently missed.
- The `grep -E` patterns now escape `+` as well as `.`.

Verified a beta-to-beta run end to end (iOS `4.7.2-beta.1` → `4.8.0-beta.2`, Android `4.20.1+1` → `4.21.0+2`): all five files updated, exactly one line changed in the pbxproj. Re-checked the stable path for regressions.

One thing I deliberately left alone: that pbxproj sed is unscoped, so if a second remote SPM package is ever added to the sample app it would rewrite that pin too. Today there's a single `version =` key and a single `XCRemoteSwiftPackageReference`, and this predates the PR — happy to scope it to the customerio-ios block if you'd rather close that off now.

PR #394, opened 2026-08-17
feat: add Flutter UIScene lifecycle adapter (MBL-2277)

Shahroz16:
## Summary

Adds the production Flutter UIScene adapter needed for Live Activity tap attribution and redirect routing on iOS 27.

- AppDelegate remains responsible for SDK initialization, APNs registration, and the global notification-center delegate.
- The scene adapter handles only Customer.io Live Activity URLs. Ordinary URLs, user activities, notification responses, and shortcuts remain with Flutter or the host.
- Opened metrics and redirect delivery are deduplicated per UIKit URL occurrence when Flutter offers a callback to multiple engines.
- Redirects use the Flutter view controller associated with the callback's scene and a bounded cold-start wait matching Flutter's three-second first-frame timeout.
- The plugin uses released native iOS SDK 4.7.5 through CocoaPods and SwiftPM. No branch or commit pin remains.
- Release tooling pins semantic-release 25.0.9 with conventionalcommits 9.3.1 to prevent dependency drift from breaking deployment.

## Compatibility

- AppDelegate-only apps keep the existing manual `CustomerIOLiveActivities.handleWidgetUrl` integration.
- UIScene apps using Flutter deep linking get automatic attribution and redirect routing.
- Apps with `FlutterDeepLinkingEnabled=false` keep host lifecycle ownership and the manual handler.
- Scene registration is dynamic so older AppDelegate-only Flutter hosts still compile. UIScene integration requires Flutter 3.44.8 or newer.
- Custom multi-scene engine topologies keep Flutter's documented scene-to-engine registration in the host scene delegate.

This adapter does not register an application lifecycle delegate or change notification, APN, or FCM ownership. Multiple-push-provider composition comes from the released native iOS SDK.

## Migration

- AppDelegate-only hosts: no change.
- UIScene hosts using Flutter deep linking: keep `CioAppDelegateWrapper` and remove any manual Customer.io URL call from the host `SceneDelegate` before upgrading. The plugin now owns that Live Activity path automatically; keeping both can double-report opens.
- UIScene hosts with `FlutterDeepLinkingEnabled=false`: keep the manual host lifecycle handler.
- Custom multi-scene engine topologies: keep the host's Flutter scene-to-engine registration.

The merged iOS 4.7.5 and Android 4.20.2 dependency updates require no migration.

## Scope

This PR contains production routing, a focused routing behavior test, sample/CI wiring, released dependency pins, and customer-facing lifecycle guidance. It contains no custom lifecycle schema, source lock, trace recorder, capture runner, or cross-repository validation framework.

## Validation

- Flutter 3.44.8 root tests: 111/111 passed.
- Direct Swift routing and occurrence tests: passed.
- Lint, API checks, publish validation, security checks, and Cursor Bugbot: passed on the final head.
- SwiftPM and CocoaPods UIScene builds, simulator launch/survival, and Flutter scene callback checks: [passed](https://github.com/customerio/customerio-flutter/actions/runs/32498670311) on lifecycle head `7c59797`, which is unchanged by the subsequent release-workflow pin.
- `git diff --check`: passed.

Fresh Claude Code Opus reviews challenged the complete branch against `origin/main` and the final exact head. Valid findings were implemented, including wrong-scene engine selection, cold-start redirect readiness, duplicate per-occurrence delivery, nested tracking redirects, removal of an impossible optional navigation-channel guard, and cache identity verification. Suggestions that would reintroduce source-shape tests, production logs used only as CI markers, or partial handling of Flutter's all-or-nothing scene occurrence were not adopted.

Real-device APN/FCM delivery, terminated notification opens, and a physical Live Activity tap with one opened metric remain the explicit MBL-2233 validation boundary. This PR does not present simulator builds as device evidence.

---

> [!NOTE]
> **Medium Risk**
> Changes iOS URL and scene lifecycle routing for Live Activities, which can affect attribution and deep-link behavior, but the implementation preserves legacy AppDelegate-only behavior and is gated behind UIScene manifest detection.
> 
> **Overview**
> Adds a Flutter UIScene lifecycle adapter so iOS 27 apps using `UIApplicationSceneManifest` can automatically attribute Live Activity taps and route their redirects through Flutter's deep-link channel.
> 
> The adapter registers itself as a scene delegate only when the app declares a scene manifest and `FlutterDeepLinkingEnabled` is not explicitly disabled. It resolves Customer.io widget URLs, reports opened metrics, deduplicates per-UIKit-occurrence handling across multiple Flutter engines, and forwards valid redirects to the Flutter view controller in the callback's scene with a bounded cold-start wait. AppDelegate-only apps keep the existing manual `CustomerIOLiveActivities.handleWidgetUrl` integration unchanged.
> 
> Also adds direct Swift behavior tests for routing resolution, cold-connection ownership, and occurrence deduplication, and updates README and sample-app docs with UIScene migration guidance.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7c59797889acc9916bced34cb9f275a5debb1464. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pubspec.yaml
```diff
@@ -43,5 +43,5 @@ flutter:
         pluginClass: CustomerIOPlugin
       ios:
         pluginClass: CustomerIOPlugin
-        native_sdk_version: 4.7.2
+        native_sdk_version: 4.7.3
```
mahmoud-elmorabea: **Blocker.** 4.7.3 doesn't contain the API this PR calls.

```
$ git grep -l CioSceneLifecycleCoordinator 4.7.3 -- Sources
(nothing)
$ git merge-base --is-ancestor 09dc5fad 4.7.3
(non-zero)
```

`ios/customer_io.podspec:9` reads this exact value and turns it into `s.dependency "CustomerIO/DataPipelines", native_sdk_version` (plus MessagingInApp, MessagingInbox, and every subspec). So a customer who runs `pod install` resolves released 4.7.3 and fails to compile `CustomerIOFlutterLifecycle.swift`. `Package.swift` gets a temporary revision pin and a "replace before merge" comment; the CocoaPods path silently doesn't, and the sample masks it by force-installing the branch.

That also makes the comment above the SPM pin false — SPM resolves `09dc5fad`, CocoaPods resolves `4.7.3`, so "one plugin version always resolves one native SDK across every package manager" no longer holds.

Separately: bumping 4.7.2 → 4.7.3 here **collides with flutter#385**, which is the open auto-update PR doing exactly this bump across all four pin sites. Better to leave the version alone in this PR and let #385 own it, then set the pin to whatever release actually carries ios#1215.

ios/customer_io/Package.swift
```diff
@@ -139,7 +139,11 @@ let package = Package(
     dependencies: [
         // Pinned exactly, matching `native_sdk_version` in pubspec.yaml and the podspec, so one
         // plugin version always resolves one native SDK across every package manager.
-        .package(url: "https://github.com/customerio/customerio-ios.git", exact: "4.7.2"),
+        // Temporary stacked-PR pin. Replace with the released native SDK version before merge.
+        .package(
+            url: "https://github.com/customerio/customerio-ios.git",
+            revision: "09dc5fad234045098923863565a59b26a3ecabb5"
```
mahmoud-elmorabea: This reformat breaks `auto-update-native-sdks.yml`, and it will stay broken after you swap `revision:` back to `exact:` if the multi-line shape is kept.

The workflow does a single-line sed:

```bash
sed -i "s/\.package(url: \"https:\/\/github.com\/customerio\/customerio-ios.git\", exact: \".*\"/…/" ios/customer_io/Package.swift
```

with the URL and `exact:` on one line. With them split across lines it matches nothing, sed silently no-ops, and then `verify_updated ios/customer_io/Package.swift "customerio-ios\.git\", exact: \"$IOS_VERSION_RE\""` exits 1 and fails the run.

So the next scheduled native-SDK bump errors out instead of opening a PR. Please restore the single-line form when you drop the temporary pin — or update the sed and the assertion in the same change.

apps/flutter_sample_spm/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -967,8 +967,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/customerio/customerio-ios.git";
 			requirement = {
-				kind = exactVersion;
-				version = 4.7.2;
+				kind = revision;
+				revision = 09dc5fad234045098923863565a59b26a3ecabb5;
```
mahmoud-elmorabea: Same workflow, second failure. `kind = exactVersion; version = 4.7.2;` → `kind = revision; revision = 09dc5fad…;` leaves **no `version =` key anywhere in this pbxproj** (I grepped the file on this branch — only `kind` and `revision` match). So:

```bash
verify_updated apps/flutter_sample_spm/ios/Runner.xcodeproj/project.pbxproj "version = $IOS_VERSION_RE;"
```

exits 1 on the next bump run, and the `sed -E "s/^([[:space:]]*)version = [^;]+;/…"` above it has nothing to rewrite.

This one worries me a bit more than the Package.swift case, because per our own experience a missed SPM pin here normally fails *silently* with CI staying green. Reverting to `exactVersion` on merge fixes both — just make sure both revert together, since they're the two halves of the same invariant.

apps/flutter_sample_cocoapods/ios/Podfile
```diff
@@ -49,7 +49,8 @@ target 'Runner' do
   pod 'customer_io/liveactivities', :path => '.symlinks/plugins/customer_io/ios' # opt in to Live Activities
 
   # install_non_production_ios_sdk_local_path(local_path: '~/code/customerio-ios/', is_app_extension: false, push_service: "fcm")
-  # install_non_production_ios_sdk_git_branch(branch_name: 'feature/wrappers-inline-support', is_app_extension: false, push_service: "fcm")
+  # Review-only stack pin. Remove it and use the released coordinator version before merge.
+  install_non_production_ios_sdk_git_branch(branch_name: 'codex/mbl-2231-scene-coordinator', is_app_extension: false, push_service: "fcm")
```
mahmoud-elmorabea: Checked-in `install_non_production_ios_sdk_git_branch` pins a **moving branch** while `Package.swift` and the SPM sample pin an immutable revision. Any push to `codex/mbl-2231-scene-coordinator` changes what this sample resolves, so the two samples can compile against different native sources and a green CocoaPods build stops telling us anything about `09dc5fad`.

Worse, `get_all_cio_pods` in `customerio-ios/scripts/cocoapods_override_sdk.rb` doesn't cover the LiveActivities pods, so this Runner target links **branch** builds of DataPipelines/Common next to **trunk 4.7.3** builds of `CustomerIOLiveActivities*` — a mix no released configuration ever produces, which makes any Live Activity behavior seen in this sample unreliable as evidence.

If the override has to stay for review, use the same immutable revision the other two files use. The checked-in default should go back to the commented-out form before merge.

ios/customer_io/Sources/customer_io/Lifecycle/CustomerIOFlutterLifecycle.swift
```diff
@@ -0,0 +1,272 @@
...
+        ? CioSceneLifecycleCoordinator()
+        : nil
+
+    init(bundle: Bundle = .main) {
+        let configuredTopology = bundle.object(
+            forInfoDictionaryKey: Self.hostTopologyInfoPlistKey
+        ) as? String
+
+        switch configuredTopology {
+        case nil:
```
mahmoud-elmorabea: This is the failure mode I'd most want closed. A host that has adopted UIScene but hasn't added `CustomerIOAppLifecycleHostTopology` falls back to `.appDelegateOnly` — and in a scene app iOS **never** calls `application(_:open:options:)`, so the seat this registers provably cannot fire.

Result: every Live Activity tap yields zero `opened` metrics and no deep-link routing. The only signal is one `DIGraphShared` error log at plugin-registration time, which nobody reads in a release build.

And this PR's README change removes the backstop: *"No AppDelegate URL override is required. Remove the earlier manual Customer.io Live Activity URL override after upgrading."* An existing UIScene app that follows that instruction and doesn't notice the new plist key goes from working attribution to silently zero.

You already read the scene manifest on line 59 to emit the log. Since you know it's a scene host, either default to `.uiScene`, or refuse the app-delegate seat entirely rather than registering one that can't work. Failing loudly (or correctly) beats registering a dead seat.

ios/customer_io/Sources/customer_io/Lifecycle/CustomerIOFlutterLifecycle.swift
```diff
@@ -0,0 +1,272 @@
...
+                DIGraphShared.shared.logger.error(
+                    "\(Self.hostTopologyInfoPlistKey) is missing for a UIScene host; " +
+                        "set it to ui-scene to enable topology-specific Customer.io routing"
+                )
+            }
+        case HostTopology.appDelegateOnly.rawValue:
+            self.hostTopology = .appDelegateOnly
+        case HostTopology.uiScene.rawValue:
+            self.hostTopology = .uiScene
+        default:
```
mahmoud-elmorabea: `self.hostTopology = nil` on an unrecognized value turns *both* seats off: `shouldRegisterApplicationDelegate` and `shouldRegisterSceneDelegate` are both false, and `handleApplicationOpenURL` also bails on its `hostTopology == .appDelegateOnly` guard.

So `<string>uiscene</string>`, `<string>UIScene</string>`, or a trailing space costs you all Live Activity attribution for the life of that build — again with one log line, and again with the manual override removed per the README.

Falling back to `.appDelegateOnly` (today's behavior on `main`) keeps attribution working for the legacy case while still surfacing the misconfiguration in the log. Nil is the one choice with no working path.

ios/customer_io/Sources/customer_io/Lifecycle/CustomerIOFlutterLifecycle.swift
```diff
@@ -0,0 +1,272 @@
...
+            guard hostTopology == .appDelegateOnly else { return false }
+            return routeURL(url, applicationOptions: options)
+        }
+    }
+
+    @available(iOS 13.0, *)
+    func handleSceneConnection(_ connectionOptions: UIScene.ConnectionOptions?) -> Bool {
+        return MainActor.assumeIsolated {
+            guard let sceneCoordinator, let connectionOptions else { return false }
+            guard connectionOptions.urlContexts.count == 1,
```
mahmoud-elmorabea: This re-implements the coordinator's ambiguity policy, more strictly, and drops the metric before `handleWidgetUrl` is ever called — with no log.

`handleSceneConnection` requires exactly one URL context and no user activity or shortcut, then calls `sceneCoordinator.handleConnection`, which computes the identical `candidateCount <= 1` check. `handleSceneOpenURLContexts:133` does the same for the warm path — even though `CioSceneLifecycleCoordinator.handleOpenURLs` exists specifically to iterate and route *every* URL once. So when iOS coalesces two opens into one `scene:openURLContexts:` and one of them is the `cio-live-activity://` tap, the tap is invisible in reporting.

Two consequences worth separating:

1. **Two copies of the ambiguity rule** that can drift apart. I raised the same policy question on customerio-ios#1215 — the URL candidates arguably shouldn't be suppressed at all, since reporting a delivery is never ambiguous even when *navigation* is. Whatever we decide, it should live in the coordinator only, and this adapter should just forward.
2. **The drop should at minimum be logged.** "Reporting silently disagrees with what the user did" is exactly the class of bug that takes months to notice.

Also, the closures passed for `continueUserActivity:` and `performShortcut:` are provably unreachable given the guard above, which makes `.rejectedAmbiguousInput` unreachable in `claim()` too.

ios/customer_io/Sources/customer_io/Lifecycle/CustomerIOFlutterLifecycle.swift
```diff
@@ -0,0 +1,272 @@
...
+            guard !isForwardingRedirectToApplication else { return false }
+            guard hostTopology == .appDelegateOnly else { return false }
+            return routeURL(url, applicationOptions: options)
+        }
+    }
+
+    @available(iOS 13.0, *)
+    func handleSceneConnection(_ connectionOptions: UIScene.ConnectionOptions?) -> Bool {
+        return MainActor.assumeIsolated {
+            guard let sceneCoordinator, let connectionOptions else { return false }
```
mahmoud-elmorabea: `guard let sceneCoordinator, let connectionOptions else { return false }` makes cold-start attribution depend on plugin registration order, silently.

`FlutterSceneLifeCycle.h` documents `connectionOptions` as possibly nil "if another plugin has already handled the connection," and `FlutterPluginSceneLifeCycleDelegate` calls plugins "until a plugin handles the request." So if any plugin registered before `customer_io` by `GeneratedPluginRegistrant` — a dynamic-links or deep-link plugin, say — claims the connection first, this is invoked with nil, returns false, reports no `opened`, and logs nothing.

Cold-start Live Activity attribution then works or doesn't as a function of alphabetical plugin registration order, with no diagnostic to tell us which happened. At minimum log the nil case so it's diagnosable from a customer's log bundle.

ios/customer_io/Sources/customer_io/Lifecycle/CustomerIOFlutterLifecycle.swift
```diff
@@ -0,0 +1,272 @@
...
+            DIGraphShared.shared.logger.error(
+                "Customer.io Live Activity redirect cannot target another tracking URL"
+            )
+            return false
+        }
+
+        if let appDelegate = UIApplication.shared.delegate {
+            isForwardingRedirectToApplication = true
+            defer { isForwardingRedirectToApplication = false }
+            if appDelegate.application?(
```
mahmoud-elmorabea: Two things in this block, and one is a question I couldn't settle from this repo.

**The question:** in the `.uiScene` topology, does forwarding through `appDelegate.application?(_:open:options:)` do anything? My understanding is that Flutter's `FlutterPluginAppLifeCycleDelegate` only holds plugins registered via `addApplicationDelegate:`, and in a scene app plugins register via `addSceneDelegate:` while Dart deep-link delivery runs on the scene path — which would make this return `NO` every time. The code then falls through to `UIApplication.shared.open(destination)` on line 205, bouncing the customer's deep link out to iOS and back in as a brand-new `UIOpenURLContext` (or into Safari for an `https` destination we don't claim as a universal link). Meanwhile line 211 returns `true` unconditionally, including when the async `open` reports failure, so the callback is claimed and the link is lost with only a log.

If you've verified Flutter does forward to the app delegate in a scene app, disregard — but then the README's "first passes it through the existing Flutter and host URL handlers" needs to be true on both topologies, and right now I can't tell that it is.

**The smaller one:** `defer { isForwardingRedirectToApplication = false }` clears rather than restores. If a forwarded URL ever synchronously re-enters `routeURL`, the inner `defer` clears the outer frame's guard and `handleApplicationOpenURL` stops declining — reintroducing exactly the double-handling the flag exists to prevent. `let previous = isForwardingRedirectToApplication; defer { isForwardingRedirectToApplication = previous }`, or a depth counter.

ios/customer_io/Sources/customer_io/Lifecycle/CustomerIOFlutterLifecycle.swift
```diff
@@ -0,0 +1,272 @@
...
+        }
+    }
+
+    @MainActor
+    private static func handleSceneOccurrence(
+        _ occurrence: UIOpenURLContext,
+        route: () -> Bool
+    ) -> Bool {
+        sceneOccurrenceResults = sceneOccurrenceResults.filter { $0.value.occurrence != nil }
+        let identifier = ObjectIdentifier(occurrence)
```
mahmoud-elmorabea: The cross-engine dedup key comes from a *different read* of `connectionOptions.urlContexts` than the object the coordinator actually routes.

`handleSceneConnection` reads `urlContexts` on line 104 (`.count`), again on line 107 (`.first` → the dedup key), and `CioSceneLifecycleCoordinator.handleConnection` reads it a third time via `Array(connectionOptions.urlContexts)` to get the object passed to `routeURL`. The once-only guarantee therefore rests on UIKit vending identical `UIOpenURLContext` instances across three property accesses, which isn't documented anywhere I can find.

If it doesn't hold, with two engines registered the second computes a different `ObjectIdentifier`, misses the cache, and calls `handleWidgetUrl` again — a duplicate `opened` metric and a duplicate host forward for one tap. Reading `urlContexts` once into a local and passing that same object to both the dedup and the coordinator removes the assumption entirely.

ios/customer_io/Sources/customer_io/CustomerIOPlugin.swift
```diff
@@ -41,6 +54,42 @@ public class CustomerIOPlugin: NSObject, FlutterPlugin {
...
+    ) {
+        guard instance.lifecycleHandler.shouldRegisterSceneDelegate else { return }
+        guard #available(iOS 13.0, *) else { return }
+
+        let selector = NSSelectorFromString("addSceneDelegate:")
+        guard registrar.responds(to: selector) else {
+            instance.lifecycleHandler.reportUnavailableSceneRegistration()
+            return
+        }
+        _ = registrar.perform(selector, with: instance.lifecycleHandler)
```
mahmoud-elmorabea: Retention asymmetry between the two seats. Lines 11-19 explain that `applicationLifecycleRegistrationOwner` exists because "Flutter stores application lifecycle delegates weakly" — good. But this passes `instance.lifecycleHandler`, owned only by `instance`, and in `.uiScene` topology `applicationLifecycleRegistrationOwner` is never set.

So after `register(with:)` returns, the only thing keeping `instance` alive is the incidental retain cycle created below by `addMethodCallDelegate` (channel → handler block → instance → channel). If that cycle is ever broken — `deinit`'s `setMethodCallHandler(nil)`, an engine teardown path, or a Flutter change to how method-call delegates are held — and Flutter's scene registry is weak like the app-delegate one, all UIScene URL routing silently stops.

Make it explicit either way: retain the scene owner the same way, or register `instance` for both seats. Relying on an unrelated retain cycle is the kind of thing that survives review and breaks on a Flutter upgrade.

ios/customer_io/Sources/customer_io/CustomerIOPlugin.swift
```diff
@@ -11,6 +11,16 @@ import CioLocationGeofence
 #endif
 
 public class CustomerIOPlugin: NSObject, FlutterPlugin {
+    private static let lifecycleRegistrationLock = NSLock()
+    // UIApplication callbacks are global, so exactly one plugin instance owns that registration.
+    // Retaining that owner for the process lifetime is intentional because Flutter stores
+    // application lifecycle delegates weakly.
+    // Scene callbacks are registered per engine and deduplicated by their UIKit occurrence object
+    // in CustomerIOFlutterLifecycle.
+    private static var applicationLifecycleRegistrationOwner: CustomerIOPlugin?
```
mahmoud-elmorabea: This latches the first instance for the whole process with no hand-off path, which bites add-to-app and multi-engine hosts.

Engine A registers first and takes the seat. Engine A is destroyed; this static keeps instance A alive forever — along with its `methodChannel` bound to a dead binary messenger, `inAppMessagingChannelHandler`, and `liveActivitiesChannelHandler` — sitting in Flutter's weak delegate list. Engine B then registers, sees the owner is non-nil, and skips registration. URL routing is now permanently served by the dead engine's instance, and the whole plugin object graph leaks for the process lifetime.

Nothing releases the seat when the owning engine goes away. A `deinit`/detach hook that nils the owner when its engine tears down would let the next engine claim it.

test/ios27_lifecycle/production_scene_adapter_test.dart
```diff
@@ -0,0 +1,175 @@
+import 'dart:io';
+
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  final Directory root = _repoRoot();
+  String read(String path) => File('${root.path}/$path').readAsStringSync();
```
mahmoud-elmorabea: Every assertion in this file is `contains('<swift source text>')`, so it can't fail for a logic error. Flip `case .handled: return true` to `return false`, or invert `guard destination != url`, and all six tests stay green — on a 272-line file whose whole job is routing decisions. That's false confidence rather than coverage.

customerio-flutter has no Swift test target, which is the real constraint — but the native side does have `Tests/DataPipeline/Integration/CioSceneLifecycleCoordinatorTests.swift`, so the behavioral assertions belong there (and I've asked on ios#1215 for the public entry points to be covered too). Here I'd keep only what pins a contract we'd otherwise lose silently, and drop the source-substring checks.

Two smaller things:

- `read()` is re-implemented here even though `flutter_scene_fixture_test.dart` in this same directory already defines a top-level `read(String path)` — and this copy drops that one's `existsSync` assertion, so a moved file reads as empty rather than failing loudly. `test/messaging_in_app/test_helpers.dart` shows the repo's shared-helper pattern.
- `expect(legacy, contains('<string>app-delegate-only</string>'))` (line 138, and `flutter_scene_fixture_test.dart:72`) asserts the *value* with no reference to the `CustomerIOAppLifecycleHostTopology` key. Given the two findings above about a missing or misspelled key silently disabling attribution, this is the one assertion I'd most want to be key-aware.

ios/customer_io/Package.swift
```diff
@@ -139,7 +139,11 @@ let package = Package(
     dependencies: [
         // Pinned exactly, matching `native_sdk_version` in pubspec.yaml and the podspec, so one
         // plugin version always resolves one native SDK across every package manager.
-        .package(url: "https://github.com/customerio/customerio-ios.git", exact: "4.7.2"),
+        // Temporary stacked-PR pin. Replace with the released native SDK version before merge.
+        .package(
+            url: "https://github.com/customerio/customerio-ios.git",
+            revision: "e02214af7dffbb5c8ec3eb14071233d6dc55595e"
```
mahmoud-elmorabea: Thanks for refreshing this to `e02214af` — that resolves the stale-pin problem, since ios#1215 was force-pushed and `09dc5fad` no longer exists on that branch.

The remaining issue is the *shape*, not the SHA: this reformat breaks `auto-update-native-sdks.yml`, and it stays broken after you swap `revision:` back to `exact:` if the multi-line form is kept.

The workflow does a single-line sed:

```bash
sed -i "s/\\.package(url: \\"https:\\/\\/github.com\\/customerio\\/customerio-ios.git\\", exact: \\".*\\"/…/" ios/customer_io/Package.swift
```

with the URL and `exact:` on one line. Split across lines it matches nothing, sed silently no-ops, and `verify_updated ios/customer_io/Package.swift "customerio-ios\\.git\\", exact: \\"$IOS_VERSION_RE\\""` then exits 1 — so the next scheduled bump run fails instead of opening a PR.

Please restore the single-line form when the temporary pin comes out, or update the sed and its assertion in the same change.

apps/flutter_sample_spm/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -967,8 +967,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/customerio/customerio-ios.git";
 			requirement = {
-				kind = exactVersion;
-				version = 4.7.2;
+				kind = revision;
+				revision = e02214af7dffbb5c8ec3eb14071233d6dc55595e;
```
mahmoud-elmorabea: Same workflow, second failure — and this one is the more dangerous of the two.

`kind = exactVersion; version = 4.7.2;` became `kind = revision; revision = <sha>;`, so there is **no `version =` key left anywhere in this pbxproj** (grepped on the current head — only `kind` and `revision` match). That means:

```bash
verify_updated apps/flutter_sample_spm/ios/Runner.xcodeproj/project.pbxproj "version = $IOS_VERSION_RE;"
```

exits 1 on the next bump, and the `sed -E "s/^([[:space:]]*)version = [^;]+;/…"` above it has nothing to rewrite.

I flag this one harder because a missed SPM pin here normally fails *silently* with CI staying green, so this file is exactly where we can least afford the bump automation to stop working. Reverting to `exactVersion` on merge fixes both halves — just make sure it happens together with `Package.swift`, since they're two halves of one invariant.

apps/flutter_sample_cocoapods/ios/Podfile
mahmoud-elmorabea: Checked-in `install_non_production_ios_sdk_git_branch` pins a **moving branch** while `Package.swift` and the SPM sample pin an immutable revision. Any push to `codex/mbl-2231-scene-coordinator` changes what this sample resolves, so the two samples can compile against different native sources and a green CocoaPods build stops telling us anything about `09dc5fad`.

Worse, `get_all_cio_pods` in `customerio-ios/scripts/cocoapods_override_sdk.rb` doesn't cover the LiveActivities pods, so this Runner target links **branch** builds of DataPipelines/Common next to **trunk 4.7.3** builds of `CustomerIOLiveActivities*` — a mix no released configuration ever produces, which makes any Live Activity behavior seen in this sample unreliable as evidence.

If the override has to stay for review, use the same immutable revision the other two files use. The checked-in default should go back to the commented-out form before merge.

ios/customer_io/Sources/customer_io/Lifecycle/CustomerIOFlutterLifecycle.swift
mahmoud-elmorabea: This is the failure mode I'd most want closed. A host that has adopted UIScene but hasn't added `CustomerIOAppLifecycleHostTopology` falls back to `.appDelegateOnly` — and in a scene app iOS **never** calls `application(_:open:options:)`, so the seat this registers provably cannot fire.

Result: every Live Activity tap yields zero `opened` metrics and no deep-link routing. The only signal is one `DIGraphShared` error log at plugin-registration time, which nobody reads in a release build.

And this PR's README change removes the backstop: *"No AppDelegate URL override is required. Remove the earlier manual Customer.io Live Activity URL override after upgrading."* An existing UIScene app that follows that instruction and doesn't notice the new plist key goes from working attribution to silently zero.

You already read the scene manifest on line 59 to emit the log. Since you know it's a scene host, either default to `.uiScene`, or refuse the app-delegate seat entirely rather than registering one that can't work. Failing loudly (or correctly) beats registering a dead seat.

ios/customer_io/Sources/customer_io/Lifecycle/CustomerIOFlutterLifecycle.swift
mahmoud-elmorabea: This re-implements the coordinator's ambiguity policy, more strictly, and drops the metric before `handleWidgetUrl` is ever called — with no log.

`handleSceneConnection` requires exactly one URL context and no user activity or shortcut, then calls `sceneCoordinator.handleConnection`, which computes the identical `candidateCount <= 1` check. `handleSceneOpenURLContexts:133` does the same for the warm path — even though `CioSceneLifecycleCoordinator.handleOpenURLs` exists specifically to iterate and route *every* URL once. So when iOS coalesces two opens into one `scene:openURLContexts:` and one of them is the `cio-live-activity://` tap, the tap is invisible in reporting.

Two consequences worth separating:

1. **Two copies of the ambiguity rule** that can drift apart. I raised the same policy question on customerio-ios#1215 — the URL candidates arguably shouldn't be suppressed at all, since reporting a delivery is never ambiguous even when *navigation* is. Whatever we decide, it should live in the coordinator only, and this adapter should just forward.
2. **The drop should at minimum be logged.** "Reporting silently disagrees with what the user did" is exactly the class of bug that takes months to notice.

Also, the closures passed for `continueUserActivity:` and `performShortcut:` are provably unreachable given the guard above, which makes `.rejectedAmbiguousInput` unreachable in `claim()` too.

ios/customer_io/Sources/customer_io/Lifecycle/CustomerIOFlutterLifecycle.swift
mahmoud-elmorabea: `guard let sceneCoordinator, let connectionOptions else { return false }` makes cold-start attribution depend on plugin registration order, silently.

`FlutterSceneLifeCycle.h` documents `connectionOptions` as possibly nil "if another plugin has already handled the connection," and `FlutterPluginSceneLifeCycleDelegate` calls plugins "until a plugin handles the request." So if any plugin registered before `customer_io` by `GeneratedPluginRegistrant` — a dynamic-links or deep-link plugin, say — claims the connection first, this is invoked with nil, returns false, reports no `opened`, and logs nothing.

Cold-start Live Activity attribution then works or doesn't as a function of alphabetical plugin registration order, with no diagnostic to tell us which happened. At minimum log the nil case so it's diagnosable from a customer's log bundle.

ios/customer_io/Sources/customer_io/Lifecycle/CustomerIOFlutterLifecycle.swift
mahmoud-elmorabea: Two things in this block, and one is a question I couldn't settle from this repo.

**The question:** in the `.uiScene` topology, does forwarding through `appDelegate.application?(_:open:options:)` do anything? My understanding is that Flutter's `FlutterPluginAppLifeCycleDelegate` only holds plugins registered via `addApplicationDelegate:`, and in a scene app plugins register via `addSceneDelegate:` while Dart deep-link delivery runs on the scene path — which would make this return `NO` every time. The code then falls through to `UIApplication.shared.open(destination)` on line 205, bouncing the customer's deep link out to iOS and back in as a brand-new `UIOpenURLContext` (or into Safari for an `https` destination we don't claim as a universal link). Meanwhile line 211 returns `true` unconditionally, including when the async `open` reports failure, so the callback is claimed and the link is lost with only a log.

If you've verified Flutter does forward to the app delegate in a scene app, disregard — but then the README's "first passes it through the existing Flutter and host URL handlers" needs to be true on both topologies, and right now I can't tell that it is.

**The smaller one:** `defer { isForwardingRedirectToApplication = false }` clears rather than restores. If a forwarded URL ever synchronously re-enters `routeURL`, the inner `defer` clears the outer frame's guard and `handleApplicationOpenURL` stops declining — reintroducing exactly the double-handling the flag exists to prevent. `let previous = isForwardingRedirectToApplication; defer { isForwardingRedirectToApplication = previous }`, or a depth counter.

ios/customer_io/Sources/customer_io/Lifecycle/CustomerIOFlutterLifecycle.swift
mahmoud-elmorabea: The cross-engine dedup key comes from a *different read* of `connectionOptions.urlContexts` than the object the coordinator actually routes.

`handleSceneConnection` reads `urlContexts` on line 104 (`.count`), again on line 107 (`.first` → the dedup key), and `CioSceneLifecycleCoordinator.handleConnection` reads it a third time via `Array(connectionOptions.urlContexts)` to get the object passed to `routeURL`. The once-only guarantee therefore rests on UIKit vending identical `UIOpenURLContext` instances across three property accesses, which isn't documented anywhere I can find.

If it doesn't hold, with two engines registered the second computes a different `ObjectIdentifier`, misses the cache, and calls `handleWidgetUrl` again — a duplicate `opened` metric and a duplicate host forward for one tap. Reading `urlContexts` once into a local and passing that same object to both the dedup and the coordinator removes the assumption entirely.

ios/customer_io/Sources/customer_io/CustomerIOPlugin.swift
mahmoud-elmorabea: This latches the first instance for the whole process with no hand-off path, which bites add-to-app and multi-engine hosts.

Engine A registers first and takes the seat. Engine A is destroyed; this static keeps instance A alive forever — along with its `methodChannel` bound to a dead binary messenger, `inAppMessagingChannelHandler`, and `liveActivitiesChannelHandler` — sitting in Flutter's weak delegate list. Engine B then registers, sees the owner is non-nil, and skips registration. URL routing is now permanently served by the dead engine's instance, and the whole plugin object graph leaks for the process lifetime.

Nothing releases the seat when the owning engine goes away. A `deinit`/detach hook that nils the owner when its engine tears down would let the next engine claim it.

test/ios27_lifecycle/production_scene_adapter_test.dart
mahmoud-elmorabea: Every assertion in this file is `contains('<swift source text>')`, so it can't fail for a logic error. Flip `case .handled: return true` to `return false`, or invert `guard destination != url`, and all six tests stay green — on a 272-line file whose whole job is routing decisions. That's false confidence rather than coverage.

customerio-flutter has no Swift test target, which is the real constraint — but the native side does have `Tests/DataPipeline/Integration/CioSceneLifecycleCoordinatorTests.swift`, so the behavioral assertions belong there (and I've asked on ios#1215 for the public entry points to be covered too). Here I'd keep only what pins a contract we'd otherwise lose silently, and drop the source-substring checks.

Two smaller things:

- `read()` is re-implemented here even though `flutter_scene_fixture_test.dart` in this same directory already defines a top-level `read(String path)` — and this copy drops that one's `existsSync` assertion, so a moved file reads as empty rather than failing loudly. `test/messaging_in_app/test_helpers.dart` shows the repo's shared-helper pattern.
- `expect(legacy, contains('<string>app-delegate-only</string>'))` (line 138, and `flutter_scene_fixture_test.dart:72`) asserts the *value* with no reference to the `CustomerIOAppLifecycleHostTopology` key. Given the two findings above about a missing or misspelled key silently disabling attribution, this is the one assertion I'd most want to be key-aware.

mrehan27 (approved):

PR #395, opened 2026-08-17
test: cover Flutter legacy and UIScene host lifecycles (MBL-2232)

Shahroz16:
## Summary

- add explicit legacy and UIScene sample-app topologies for SwiftPM and CocoaPods
- keep plugin registration safe for UI-less launches and Flutter implicit engines, once per engine
- prove the expected Xcode 27 legacy launch rejection and standard UIScene launch callbacks in the nightly matrix

## Scope

This is sample, test, and CI coverage only. It does not change the published Flutter plugin runtime or trigger a package release. Generic contract, capture, and routing self-test tooling was retired after review. This PR proves standard host lifecycle behavior; production URL routing remains in #394, which is stacked directly on this PR, with real-device delivery covered by MBL-2233.

## Validation

- Flutter 3.44.8 focused fixture tests: 6/6
- `flutter analyze`: no issues
- actionlint and `git diff --check`: passed
- SwiftPM and CocoaPods simulator builds: passed with Xcode 26.6
- hosted coverage includes Xcode 27 SwiftPM/CocoaPods launch checks and both stable iOS sample builds; see the PR checks for current exact-head status
- an earlier Claude Opus review motivated this simplification; the single final-head source-command-review attempt returned HTTP 429 for the individual spend limit, so no final-head Opus verdict is claimed

---

> [!NOTE]
> **Low Risk**
> Changes are limited to sample apps, tests, docs, and a non-blocking nightly workflow; no published plugin runtime behavior changes.
> 
> **Overview**
> Adds **sample-app and CI coverage** for Xcode 27’s UIScene requirement without changing the published Flutter plugin.
> 
> The **CocoaPods and SwiftPM fixtures** now support two host topologies: legacy `Info.plist` (AppDelegate-only) and `Info-Scene.plist` with `SceneDelegate` / `FlutterSceneDelegate`, selected via `CIO_LIFECYCLE_INFOPLIST_SUFFIX`. **AppDelegate** registers plugins once per engine through `FlutterImplicitEngineDelegate` and a registry guard, removing reliance on `rootViewController` at launch.
> 
> The **Xcode 27 preview workflow** goes beyond compile-only checks: it builds the legacy app, expects launch failure with the UIScene runtime signature, then builds and launches the scene-enabled app (10s survival), and asserts scene lifecycle logs via `launch-simulator-app`. Ruby is bumped to 3.4; workflow concurrency and artifact uploads are adjusted. **Docs** and **`flutter_scene_fixture_test.dart`** lock in these invariants.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1bdc9b8cd49dea3594ddac65b05f7723b83449b3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Shahroz16:
Final CI follow-up: the Xcode preview lane was the repository straggler on Ruby 3.1, while the locked connection_pool 3.0.2 requires Ruby 3.2+ and regular sample-build CI already uses Ruby 3.4. The preview lane now uses the same Ruby 3.4 standard. Claude Opus returned ACCEPT with no lockfile or compatibility concern. The fresh hosted job has passed locked Ruby installation and is compiling the CocoaPods fixture.

Shahroz16:
Addressed both review points in 22ea5c9:

- Restored launch-time plugin registration through FlutterAppDelegate for UI-less remote-notification/background-fetch wakes, while retaining didInitializeImplicitFlutterEngine for callback-created or additional engines behind one registry-key guard.
- Updated the preview lane to ruby/setup-ruby v1.321.0 with Ruby 3.4.
- Replaced compile-only evidence with an Xcode 27 negative control for the AppDelegate-only product and a positive install/launch/10-second survival check for the UIScene product, for both SwiftPM and CocoaPods.

The shared launch action is still a draft dependency, so this PR must not merge until customerio/mobile-ci-tools #16 is merged and the workflow reference is switched to @main. Hosted Xcode 27 checks are running now.

Shahroz16:
Hosted Xcode 27 validation is complete and green for both SwiftPM and CocoaPods. In each matrix job, the AppDelegate-only control compiled and then failed with the expected UIScene launch signature; the UIScene-enabled product installed, launched, and kept the expected executable alive for 10 seconds. Run: https://github.com/customerio/customerio-flutter/actions/runs/32054303030\n\nMerge dependency is unchanged: mobile-ci-tools #16 must merge first, then this workflow reference must be switched to `@main` and rerun.

Shahroz16:
Follow-up review corrections are now pushed in 14adff7. I verified Flutter 3.44.8 engine behavior directly: FlutterAppDelegate uses its launch engine before a root controller exists, and the storyboard FlutterViewController later adopts that same engine before the implicit-engine callback. The registry-key guard therefore prevents duplicate generated registration on the primary engine while preserving registration for a distinct later engine. FlutterAppDelegate URL forwarding also targets the root FlutterViewController engine. The valid review findings are addressed: per-engine permission-handler retention, one retained scene handler, scene-handler registration before generated plugins, an explicit engine-readiness gate, a 10-second bounded retry, and less brittle active-callback CI. Focused tests pass 6/6 and actionlint/diff checks pass. The final exact-diff Opus retry was unavailable after the bounded wait and returned only Execution error, so I am not representing that retry as an acceptance; the earlier substantive Opus review produced the findings above, which were independently adjudicated and corrected. Exact-head hosted builds are now running.

apps/flutter_sample_spm/ios/Runner/SceneDelegate.swift
```diff
@@ -0,0 +1,30 @@
...
+    )
+    private let runToken = ProcessInfo.processInfo.environment["CIO_SCENE_HANDLER_RUN_TOKEN"] ?? "none"
+
+    override func scene(
+        _ scene: UIScene,
+        willConnectTo session: UISceneSession,
+        options connectionOptions: UIScene.ConnectionOptions
+    ) {
+        logger.notice(
+            "customerio-flutter-scene-will-connect token=\(self.runToken, privacy: .public)"
```
mahmoud-elmorabea: **This is the finding I'd most want addressed.** The `notice` fires before `super.scene(...)`, and `sceneDidBecomeActive` does the same on line 26.

So the two greps in `Verify Flutter scene callbacks` prove exactly one thing: UIKit called this subclass override. They say nothing about whether `FlutterSceneDelegate` forwarded the callback onward to `sceneLifeCycleDelegate`, the engine, and the registered plugins.

If a Flutter engine or Xcode 27 regression breaks that forwarding — plugins never see `scene:willConnectToSession:`, the engine never activates — these lines still land in the log store, the greps pass, and the job reports green with `**Scope:** ... standard Flutter scene callback forwarding` and the docs claim it proves "delivery ... through Flutter's standard scene forwarding."

Moving the logs after `super` is a small improvement (it at least proves `super` returned), but it still doesn't observe the far side. The assertion needs to key off something the engine or a plugin produces — a Dart-side lifecycle notice over a method channel, or a plugin's own `os_log` — so a forwarding break actually turns the nightly red.

.github/workflows/ios-toolchain-compatibility.yml
```diff
@@ -94,15 +94,101 @@ jobs:
...
+            fi
+            sleep 1
+          done
+          grep -Fq "customerio-flutter-scene-will-connect token=$run_token" "$callback_log"
+          grep -Fq "customerio-flutter-scene-did-become-active token=$run_token" "$callback_log"
+          test "$(grep -Fc "customerio-flutter-scene-will-connect token=$run_token" "$callback_log")" -eq 1
+          test "$(grep -Fc "customerio-flutter-scene-did-become-active token=$run_token" "$callback_log")" -ge 1
+
+      - name: Record launch result
+        if: steps.scene-launch.outcome == 'success'
```
mahmoud-elmorabea: An explicit `if:` replaces GitHub's implicit `success()`, so this step still runs when a *later* step in the same job has failed.

Concretely: `scene-launch` succeeds, `Verify Flutter scene callbacks` fails on a missing will-connect line. The job goes red, but `steps.scene-launch.outcome == 'success'` is still true, so this appends `## Flutter fixture launch` / `**Classification:** launch-passed` and a `**Scope:**` line claiming "standard Flutter scene callback forwarding" — which the run just disproved — directly above the `scene-callback-verification-failed` section from line 246. Whoever triages the nightly reads a pass and a fail for the same run.

`if: success() && steps.scene-launch.outcome == 'success'` fixes it.

docs/ios-toolchain-compatibility-ci.md
```diff
@@ -1,15 +1,15 @@
...
 
-Preview failures are recorded as failed scheduled or manually dispatched runs. They cannot block ordinary pull requests because those pull requests do not trigger this workflow. A hosted preview label can become unavailable before a job starts, and job timeouts do not cover queue time, so a missing or persistently queued nightly is an infrastructure alert rather than a pass.
+Preview failures are recorded as failed scheduled or manually dispatched runs. They cannot block ordinary pull requests because those pull requests do not trigger this workflow. The native iOS repository owns a separate [Ubuntu watchdog](https://github.com/customerio/customerio-ios/blob/main/.github/workflows/xcode27-nightly-watchdog.yml) that checks the iOS, Flutter, and Expo scheduled results after their normal completion window and alerts the Mobile team for a failed, incomplete, or missing run. A hosted preview label can become unavailable before a job starts, and job timeouts do not cover queue time, so a missing or persistently queued nightly is an infrastructure alert rather than a pass.
```
mahmoud-elmorabea: The linked workflow doesn't exist. `GET repos/customerio/customerio-ios/contents/.github/workflows/xcode27-nightly-watchdog.yml` → 404, and grepping that repo's whole workflow listing for `watch|nightly` returns nothing. customerio-ios#1216 (`ci: verify Xcode 27 nightly launches`) only touches `ios-toolchain-compatibility.yml` and a dev note.

That matters beyond a dead link, because the rest of the paragraph relies on the watchdog to argue that a missing or persistently-queued nightly is "an infrastructure alert rather than a pass." As written, a reader concludes the nightly is monitored when a silently-missing run currently goes unnoticed — which is exactly the gap the paragraph is trying to close.

Either land the watchdog first, or describe the current state (unmonitored) and link the ticket.

.github/workflows/ios-toolchain-compatibility.yml
```diff
@@ -94,15 +94,101 @@ jobs:
...
+              > "$callback_log"
+            if grep -Fq "customerio-flutter-scene-will-connect token=$run_token" "$callback_log" \
+              && grep -Fq "customerio-flutter-scene-did-become-active token=$run_token" "$callback_log"; then
+              break
+            fi
+            sleep 1
+          done
+          grep -Fq "customerio-flutter-scene-will-connect token=$run_token" "$callback_log"
+          grep -Fq "customerio-flutter-scene-did-become-active token=$run_token" "$callback_log"
+          test "$(grep -Fc "customerio-flutter-scene-will-connect token=$run_token" "$callback_log")" -eq 1
```
mahmoud-elmorabea: `-eq 1` on will-connect, `-ge 1` on did-become-active one line down. I think the asymmetry is unintended and the strict one will flake.

The app is launched with `--terminate-running-process` on a simulator where the previous `scene-launch` session may persist, so iOS can restore the session and connect a second window scene, and a scene can disconnect/reconnect inside the 10-second collection window. Two matches → red job → `Record scene callback verification failure` reports "the app survived but did not prove Flutter scene forwarding," when forwarding in fact worked twice.

If uniqueness genuinely matters here, worth a comment saying why; otherwise match line 188.

.github/workflows/ios-toolchain-compatibility.yml
```diff
@@ -94,15 +94,101 @@ jobs:
...
+          trap cleanup EXIT
+          xcrun simctl boot "$SIMULATOR_UDID" >/dev/null 2>&1 || true
+          xcrun simctl bootstatus "$SIMULATOR_UDID" -b
+          xcrun simctl install "$SIMULATOR_UDID" "$APP_PATH"
+          run_token="${GITHUB_RUN_ID:?}-${GITHUB_RUN_ATTEMPT:?}-${{ matrix.app }}"
+          callback_start="$(date +'%Y-%m-%d %H:%M:%S')"
+          SIMCTL_CHILD_CIO_SCENE_HANDLER_RUN_TOKEN="$run_token" \
+            xcrun simctl launch --terminate-running-process "$SIMULATOR_UDID" "$BUNDLE_ID"
+          for _ in {1..10}; do
+            xcrun simctl spawn "$SIMULATOR_UDID" log show \
```
mahmoud-elmorabea: `set -euo pipefail` on line 157 defeats the retry loop. The loop exists to tolerate log entries that haven't flushed yet, but this `simctl spawn ... log show` is unguarded — if it returns non-zero even once (the simulator is still settling right after `bootstatus` on line 167, or `log` hiccups), the step exits immediately on the first iteration and `Record scene callback verification failure` blames the fixture.

The loop tolerates *missing lines* but not a *failing command*, and the failing command is the more likely transient. `> "$callback_log" || { sleep 1; continue; }` covers it.

.github/workflows/ios-toolchain-compatibility.yml
```diff
@@ -94,15 +94,101 @@ jobs:
...
+
+      - name: Verify the AppDelegate-only control failed for the UIScene requirement
+        id: verify-legacy-control
+        shell: bash
+        env:
+          LAUNCH_OUTCOME: ${{ steps.legacy-launch.outcome }}
+          LAUNCH_LOG: ${{ runner.temp }}/${{ matrix.app }}-legacy-launch.log
+        run: |
+          set -euo pipefail
+          test "$LAUNCH_OUTCOME" = 'failure'
```
mahmoud-elmorabea: `test "$LAUNCH_OUTCOME" = 'failure'` treats *any* `legacy-launch` failure as the UIScene rejection, so infrastructure failures get misreported as a broken negative control.

If the `xcode-27` runner has no iOS 27 simulator runtime, or install/boot fails, `launch-simulator-app` fails with `failure-reason=runtime-unavailable` and only that in the log. Line 137 passes, the `grep -Eq` on line 138 fails, and `Record invalid negative control` (line 260) tells the triager to "inspect the uploaded legacy launch log before trusting the scene-fixture result" — pointing at a fixture defect when the real cause is a missing runtime, which per our experience is the *most* common hosted-preview failure mode.

Worse, because this step aborts the job, `scene-launch` and `verify-scene-callbacks` never run, so we lose all scene coverage for that night with a misleading classification.

The action already exposes `steps.legacy-launch.outputs.failure-reason` — asserting it equals `did-not-survive` separates the two cases cleanly.

.github/workflows/ios-toolchain-compatibility.yml
```diff
@@ -94,15 +94,101 @@ jobs:
...
+          set -euo pipefail
+          [[ "$SIMULATOR_UDID" =~ ^[A-Za-z0-9-]+$ ]]
+          [[ "$BUNDLE_ID" =~ ^[A-Za-z0-9.-]+$ ]]
+          callback_log="$RUNNER_TEMP/${{ matrix.app }}-scene-callbacks.log"
+          cleanup() {
+            xcrun simctl terminate "$SIMULATOR_UDID" "$BUNDLE_ID" >/dev/null 2>&1 || true
+            xcrun simctl uninstall "$SIMULATOR_UDID" "$BUNDLE_ID" >/dev/null 2>&1 || true
+          }
+          trap cleanup EXIT
+          xcrun simctl boot "$SIMULATOR_UDID" >/dev/null 2>&1 || true
```
mahmoud-elmorabea: This step re-implements the simctl orchestration `ios/launch-simulator-app/v1` already owns — boot, `bootstatus -b`, install, `launch --terminate-running-process`, a retrying `log show`, and a terminate+uninstall cleanup trap.

The concrete cost: because the shared action uninstalls and shuts the simulator down on exit, this step has to **re-boot and re-install the app the previous step just launched**. That's a second full boot on a `macos` runner, and our sample-app checks already queue behind an org-wide macOS runner cap — Flutter and RN are the only consumers of it, so this is self-inflicted.

Two inputs on the shared action (`expected-log-lines` and a child-env passthrough for `SIMCTL_CHILD_*`) would give the iOS, Flutter, and Expo nightlies the same capability instead of a 35-line special case that only this workflow gets. Given customerio-ios#1216 and customerio-expo-plugin#391 are adding the same kind of launch proof right now, this seems worth doing once in `mobile-ci-tools`.

apps/flutter_sample_spm/ios/Runner/Info-Scene.plist
```diff
@@ -0,0 +1,131 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mahmoud-elmorabea: This is a hand-copied 131-line duplicate of `Info.plist`, identical apart from the scene manifest, with nothing guarding against drift.

The copies are already visibly hand-assembled — `NSSupportsLiveActivities` sits at the top of the CocoaPods twin and at line 128 here. So the next person who adds a URL scheme, a `UIBackgroundModes` entry, or an `NS...UsageDescription` to `Info.plist` will add it to one of four files and the scene topology quietly diverges from the legacy one. Neither the nightly (process survival + two log lines) nor `flutter_scene_fixture_test.dart` (which only checks the manifest, `SceneDelegate`, `fetch`, `remote-notification`) can detect that.

Cheapest fix that fits your existing test: assert in the Dart test that the scene plist equals the legacy plist plus the `UIApplicationSceneManifest` block. Then drift fails a PR instead of surviving to the nightly.

apps/flutter_sample_spm/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -282,6 +289,7 @@
 			name = Runner;
 			packageProductDependencies = (
 				78A3181F2AECB46A00862997 /* FlutterGeneratedPluginSwiftPackage */,
+				A9A0791B2FB8CCF3498BCB38 /* LiveActivities_Attributes */,
```
mahmoud-elmorabea: Two changes here look unrelated to the scene work and aren't mentioned in the PR body:

**1. `LiveActivities_Attributes` added to the Runner target.** Object `A9A0791B2FB8CCF3498BCB38` is now in `packageProductDependencies` of both `LiveActivityWidget` (line 244) and `Runner` (line 292), plus a new `PBXBuildFile` in Runner's Frameworks phase. Xcode creates one `XCSwiftPackageProductDependency` per consuming target, so the next person who opens and saves this project gets it rewritten into two objects — pbxproj churn and conflicts. It also looks unnecessary: nothing under `apps/flutter_sample_spm/ios/Runner/` references it (`CustomerIOLiveActivities` comes from the plugin via `FlutterGeneratedPluginSwiftPackage`), the sample builds without it on `main`, and the CocoaPods sample got no equivalent — so the two samples now differ for no stated reason.

**2. `SWIFT_ACTIVE_COMPILATION_CONDITIONS = "DEBUG $(inherited)"`** (line 875, and `apps/flutter_sample_cocoapods/.../project.pbxproj:993`). There is no `#if DEBUG` / `#if !DEBUG` anywhere under either sample's `ios/` — I grepped — and the project-level Debug config doesn't set it either, so nothing consumes it. Looks like leftover from the retired self-test tooling the PR body says was dropped (the test even asserts the sibling `CIO_SCENE_CONTRACT_CONDITIONS` is gone).

Both are cheap to drop, and dropping them keeps the diff honest about its scope.

test/ios27_lifecycle/flutter_scene_fixture_test.dart
```diff
@@ -0,0 +1,162 @@
...
+
+      expect(
+        occurrences(
+          projectSource,
+          r'INFOPLIST_FILE = "Runner/Info$(CIO_LIFECYCLE_INFOPLIST_SUFFIX).plist";',
+        ),
+        3,
+      );
+      expect(
+        occurrences(projectSource, 'CIO_LIFECYCLE_INFOPLIST_SUFFIX = "";'),
```
mahmoud-elmorabea: This file asserts on raw source text of files it doesn't own — pbxproj, workflow YAML, Swift — and `.github/workflows/test.yml` runs `flutter test` on every PR. That makes it a standing tax on unrelated work:

- `occurrences(projectSource, 'CIO_LIFECYCLE_INFOPLIST_SUFFIX = "";') == 3` and the `INFOPLIST_FILE` count on line 104 break the moment someone adds a fourth Runner build configuration, or Xcode rewrites the pbxproj with different quoting or ordering. A legitimate project edit fails a test that has nothing to do with it.
- `expect(source, isNot(contains('rootViewController')))` (line 48) fails on a comment that merely mentions the word.
- Six negative assertions (lines 51, 71, 96, 100, 116, 149, 150) permanently guard strings that never existed on `main`, so they can only ever fire on a false positive.
- Lines 154-158 assert occurrence counts inside the workflow YAML, which will fight every future edit to that file.

None of it proves runtime behavior — the nightly launch job does that. I'd keep the assertions that pin *contracts we'd otherwise silently lose* (the scene manifest keys, `SceneDelegate` wiring, the shared-action pin) and delete the counting and the negative assertions.

apps/flutter_sample_spm/ios/Runner/AppDelegate.swift
```diff
@@ -13,24 +14,49 @@ import CioLocationGeofence
...
+        guard !registry.hasPlugin(Self.permissionChannelPluginKey) else { return }
+        guard let registrar = registry.registrar(forPlugin: Self.permissionChannelPluginKey) else {
+            lifecycleLogger.error("Permission channel registrar unavailable; registration will retry on the next engine seat")
+            return
+        }
+
+        GeneratedPluginRegistrant.register(with: registry)
+        let permissionHandler = PermissionChannelHandler()
+        permissionHandler.register(with: registrar.messenger())
+        permissionHandlers[ObjectIdentifier(registry as AnyObject)] = permissionHandler
```
mahmoud-elmorabea: Minor: `permissionHandlers` is write-only. Nothing ever reads the dictionary declared on line 22 — the once-per-engine guard is `registry.hasPlugin(...)` on line 29, not this map. And `PermissionChannelHandler.register(with:)` calls `channel.setMethodCallHandler(handle)`, whose closure retains `self` and is retained by the engine's binary messenger, so the handler already outlives this call.

So the map's only effect is a per-engine entry that's never pruned: a stale `PermissionChannelHandler` (each owning a `CLLocationManager`) accumulates for every dead engine, keyed by an `ObjectIdentifier` the runtime can recycle. Same at `apps/flutter_sample_cocoapods/ios/Runner/AppDelegate.swift:38`. A plain stored property, or nothing, does the same job.

apps/flutter_sample_cocoapods/ios/Runner/AppDelegate.swift
```diff
@@ -13,24 +14,49 @@ import CioLocationGeofence
 @main
 class AppDelegateWithCioIntegration: CioAppDelegateWrapper<AppDelegate> {}
 
-@objc class AppDelegate: FlutterAppDelegate {
-    private let permissionHandler = PermissionChannelHandler()
+@objc class AppDelegate: FlutterAppDelegate, FlutterImplicitEngineDelegate {
```
mrehan27, line 17: Do we intentionally want both sample apps to use the new setup? If helpful, we could keep one on the older setup for a while to help validate both approaches.

mrehan27 (approved):

apps/flutter_sample_cocoapods/ios/Runner/AppDelegate.swift
Shahroz16: Yes, intentionally. The SwiftPM and CocoaPods samples are dependency-manager variants, so keeping their host code identical lets us attribute differences to the package manager rather than different lifecycle wiring. We still validate both host approaches in each sample: `Info.plist` is the AppDelegate-only legacy control and `Info-Scene.plist` is the UIScene variant. The Xcode 27 job builds both, confirms the expected legacy launch rejection, and verifies the scene-enabled app launches and receives scene callbacks. The shared registration helper also replaces the prior forced `rootViewController` path because that is unsafe for UI-less background wakes; retaining it in one sample would preserve a known defect rather than useful backward-compatibility coverage.

PR #396, opened 2026-08-18
ci: reject conditional CocoaPods deployment targets (MBL-2278)

Shahroz16:
## What changed

- Apply the qualified deployment-target fix to Flutter's package-owned helper.
- Fail before mutation for qualified `IPHONEOS_DEPLOYMENT_TARGET[...]` keys.
- Cover target, project, and xcconfig conditional settings and document the required cleanup.

## Why

Xcode can prefer a conditional target value over an unconditional override, or a new target value can lower a higher conditional project floor. The previous audit could therefore report success for the wrong effective floor.

## Validation

- Ruby helper suite: 33 tests, 190 assertions
- Flutter's package-owned helper suite verifies the behavior locally
- diff check

No cross-repository source lock, branch pin, or runtime dependency is introduced. This PR can be reviewed independently of the equivalent native helper correction.

## Release and merge policy

This CI/build-tooling PR must be **squash merged only** under its `ci:` PR title. A merge commit would expose the branch’s `fix:` commit to semantic-release and could trigger an unintended package release. The squashed change does not alter Flutter runtime behavior.

---

> [!NOTE]
> **Medium Risk**
> Changes iOS CocoaPods post-install and CI audit behavior; `pod install` can now fail on projects that previously normalized, but only when conditional deployment targets are present and the fix is intentional.
> 
> **Overview**
> **CocoaPods deployment-target normalization and CI audit now fail closed** when they see qualified `IPHONEOS_DEPLOYMENT_TARGET[...]` keys (including whitespace before `[`) on target build settings, project configurations, or parsed xcconfigs—before any project mutation or audit table output.
> 
> That blocks cases where Xcode could pick a conditional value over an unconditional override, or where adding a numeric target override could look compliant while lowering a higher conditional project floor. Errors tell integrators to replace the matrix with one unconditional numeric floor (Podfile/podspec for generated Pods targets). Docs describe the same rule for `pod install` and the standalone audit script.
> 
> The Flutter-owned helper header no longer claims byte-identical sync with other repos; behavior is covered by an expanded Ruby test suite (target/project/xcconfig, explicit-target skip paths, standalone audit).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0767f280517f8538d4897de178bcf2efa1a6797d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ios/cocoapods_deployment_target.rb
+++ b/ios/cocoapods_deployment_target.rb
@@ -181,7 +181,7 @@ module CustomerIO
               project_configuration = matching_project_configuration(project, configuration.name)
               project_value = effective_configuration_value(
                 project_configuration,
-                context: context
+                context: "#{context}, project build configuration #{configuration.name}"
               )
               effective_value = project_value
             end
--- a/ios/cocoapods_deployment_target.rb
+++ b/ios/cocoapods_deployment_target.rb
@@ -274,8 +274,9 @@ module CustomerIO
     private_class_method :configuration_file_setting
 
     def validate_no_conditional_settings!(settings, context:)
+      conditional_setting_pattern = /\A#{Regexp.escape(BUILD_SETTING)}[[:space:]]*\[/
       conditional_keys = settings.keys.map(&:to_s).select do |key|
-        key.start_with?("#{BUILD_SETTING}[")
+        key.match?(conditional_setting_pattern)
       end.sort
       return if conditional_keys.empty?
 
--- a/ios/cocoapods_deployment_target.rb
+++ b/ios/cocoapods_deployment_target.rb
@@ -282,7 +282,8 @@ module CustomerIO
       raise AuditError,
             "#{context} has conditional #{BUILD_SETTING} settings that cannot be audited deterministically: " \
             "#{conditional_keys.join(', ')}. Replace them with one numeric, unconditional #{BUILD_SETTING} " \
-            "before running this helper."
+            "before running this helper. For a generated Pods target, change the Podfile or dependency " \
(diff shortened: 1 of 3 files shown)


ios/cocoapods_deployment_target.rb
```diff
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
-# Keep this helper byte-identical across customerio-ios, customerio-flutter, and
-# customerio-reactnative so every wrapper applies the same CocoaPods contract.
+# This packaged copy is owned by customerio-flutter and covered by its local test suite.
```
mrehan27, line 3: Same comment as on iOS. A lot of this code appears to be duplicated here. We should leverage `mobile-ci-tools` repo and reuse a single copy from there, making it easier to maintain and validate across repos.

mrehan27 (approved):

ios/cocoapods_deployment_target.rb
Shahroz16: Same comment as other :)

This local copy is intentional. mobile-ci-tools is CI-only and is not available when a customer Podfile executes. Flutter and React Native must ship a package-local Ruby helper that their installed Podfiles can require without fetching another repository, and native iOS keeps its local copy as the implementation and test owner. Moving the runtime helper into mobile-ci-tools would either break that installation boundary or introduce a network/runtime dependency into customer builds. We will keep the copies versioned with their respective packages and test the same contract in each consumer.

PR #399, opened 2026-08-23
feat: route SDK deep links through Flutter scenes

Shahroz16:
## Summary

- register the native SDK deep-link callback during Flutter plugin registration for UIScene hosts using Flutter routing, including cold taps before Dart initialization
- deliver SDK push and in-app destinations, plus native-default inbox destinations when no Dart inbox listener owns the action, to the foreground/topmost Flutter engine
- preserve AppDelegate-only hosts, `FlutterDeepLinkingEnabled=false` custom routing, Live Activity attribution, and APN/FCM coexistence
- bound cold-start readiness to three seconds, use the topmost eligible Flutter UI, and fall back to `UIApplication.open` when Flutter declines or no engine becomes available
- pin every Flutter iOS integration path to the published Customer.io iOS SDK 4.7.6

## Validation

- focused Swift routing behavior tests
- `flutter analyze --no-fatal-infos`
- `flutter test` (111 passed)
- SwiftFormat lint and `git diff --check`
- clean exact-head SwiftPM iOS sample build against 4.7.6
- clean exact-head CocoaPods resolve, deployment-target audit, and iOS sample build against 4.7.6
- hosted exact-head lint, tests, API/publish checks, SwiftPM iOS sample build, and Android sample builds
- hosted CocoaPods iOS sample rerun is queued for a macOS 26 runner; the identical clean local resolve, target audit, and build passed
- fresh exact-head Claude Opus review: APPROVE
- independent exact-head review: APPROVE

Physical-device APNs/FCM delivery and tap evidence remains tracked separately in MBL-2233.

Linear: MBL-2277

---

> [!NOTE]
> **Medium Risk**
> Changes core iOS deep-link and push/in-app navigation handoff for UIScene + Flutter hosts; mis-routing could affect cold-start taps or conflict with hosts that set their own native deep-link callback.
> 
> **Overview**
> For **UIScene apps with Flutter deep linking enabled**, the iOS plugin now installs the native Customer.io **`deepLinkCallback` at plugin registration** (so cold push/in-app taps work before Dart `CustomerIO.initialize`). Destinations—including `http`/`https`—go to the **foreground/topmost Flutter engine** via `pushRouteInformation`; the plugin only falls back to **`UIApplication.open`** when Flutter reports unhandled or no engine appears within the existing ~3s readiness window.
> 
> Routing logic is refactored around shared **delivery + view-controller traversal** (`didFlutterHandle`, topmost-first search). **AppDelegate-only** hosts and **`FlutterDeepLinkingEnabled=false`** are unchanged; docs clarify this path **replaces** the native AppDelegate continuation fallback and **cannot coexist** with a custom `SDKConfigBuilder.deepLinkCallback`. Compatibility is explicitly **single window scene** only.
> 
> Pins the bundled **customerio-ios SDK to 4.7.6** (pubspec, SPM, CocoaPods samples) and adds Swift behavior tests for Flutter handle results and traversal order.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f9db5dca6fd002bf302d27e1887e0f8259ffe28c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ios/customer_io/Sources/customer_io/Lifecycle/CustomerIOFlutterDeepLinkRouter.swift
+++ b/ios/customer_io/Sources/customer_io/Lifecycle/CustomerIOFlutterDeepLinkRouter.swift
@@ -21,8 +21,8 @@ final class CustomerIOFlutterDeepLinkRouter {
     /// Delivers an SDK-triggered destination to an eligible foreground Flutter scene. If Flutter
     /// declines it or no scene becomes eligible, the destination is opened through UIKit.
     func routeSDKDeepLink(_ url: URL) {
-        DispatchQueue.main.async { [url] in
-            self.routeSDKDeepLink(url, remainingAttempts: Self.readinessAttempts)
+        DispatchQueue.main.async { [weak self, url] in
+            self?.routeSDKDeepLink(url, remainingAttempts: Self.readinessAttempts)
         }
     }
 
@@ -39,8 +39,8 @@ final class CustomerIOFlutterDeepLinkRouter {
                 openExternally(url)
                 return
             }
-            DispatchQueue.main.asyncAfter(deadline: .now() + Self.retryInterval) { [url] in
-                self.routeSDKDeepLink(url, remainingAttempts: remainingAttempts - 1)
+            DispatchQueue.main.asyncAfter(deadline: .now() + Self.retryInterval) { [weak self, url] in
+                self?.routeSDKDeepLink(url, remainingAttempts: remainingAttempts - 1)
             }
             return
         }
--- a/ios/customer_io/Sources/customer_io/Lifecycle/CustomerIOFlutterDeepLinkRouter.swift
+++ b/ios/customer_io/Sources/customer_io/Lifecycle/CustomerIOFlutterDeepLinkRouter.swift
@@ -21,8 +21,8 @@ final class CustomerIOFlutterDeepLinkRouter {
     /// Delivers an SDK-triggered destination to an eligible foreground Flutter scene. If Flutter
     /// declines it or no scene becomes eligible, the destination is opened through UIKit.
     func routeSDKDeepLink(_ url: URL) {
-        DispatchQueue.main.async { [url] in
-            self.routeSDKDeepLink(url, remainingAttempts: Self.readinessAttempts)
+        DispatchQueue.main.async { [weak self, url] in
+            self?.routeSDKDeepLink(url, remainingAttempts: Self.readinessAttempts)
         }
     }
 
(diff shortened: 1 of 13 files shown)


ios/customer_io/Sources/customer_io/Lifecycle/CustomerIOFlutterLifecycle.swift
```diff
@@ -21,6 +21,10 @@ final class CustomerIOFlutterLifecycle: NSObject {
         #endif
     }
 
+    var shouldRegisterSDKDeepLinkCallback: Bool {
```
mahmoud-elmorabea, line 24: 🔴 The `usesUIScene` half defeats the purpose. AppDelegate-only Flutter hosts don't register and keep using the SDK's app-delegate `continue:` handoff, which only works because `FlutterAppDelegate` implements `application(_:continue:)`. They break when we remove it — the thing this PR prevents. `connectedScenes` is populated on iOS 13+ without a manifest, so the router already works for them. Can this drop to `usesFlutterDeepLinking` alone?

ios/customer_io/Sources/customer_io/Lifecycle/CustomerIOFlutterDeepLinkRouter.swift
```diff
@@ -60,11 +90,80 @@ final class CustomerIOFlutterDeepLinkRouter {
             (useHiddenFallback ? candidates.first : nil)
     }
 
+    @MainActor
+    private func applicationFlutterViewController(
+        useHiddenFallback: Bool
+    ) -> FlutterViewController? {
+        let foregroundScenes = UIApplication.shared.connectedScenes
```
mahmoud-elmorabea: 🟠 Walks `UIApplication.shared.connectedScenes` — the pattern Apple's modernization guidance calls out ("never walk global scene/window state"). Not asking you to change it: there's no scene in a `deepLinkCallback`, and `registrar?.viewController` can't handle multi-engine. Yours is deterministic (sorted by `persistentIdentifier`, then activation state, then key/displaying), which is better than native's `ModalViewManager.getUIWindow()`. Flagging so it's a recorded decision — the same pattern now lives in two repos at different quality.

ios/customer_io/Sources/customer_io/Lifecycle/CustomerIOFlutterLifecycle.swift
Shahroz16: The premise does not match the current compatibility decision. Native iOS is retaining the callback → AppDelegate `continue` → system-open fallback for legacy AppDelegate hosts. Registering the Flutter callback there would bypass existing AppDelegate filtering and change current behavior. #399 intentionally limits the new bridge to UIScene hosts with Flutter deep linking enabled. If native iOS ever removes that fallback, it will require a coordinated wrapper migration; that removal is not part of this iOS 27 work.

ios/customer_io/Sources/customer_io/Lifecycle/CustomerIOFlutterDeepLinkRouter.swift
Shahroz16: Agreed. The updated README now records the boundary explicitly: the compatibility scope is one simultaneous window scene, and this deterministic `connectedScenes` selection is defensive behavior rather than a multi-window support guarantee. I have not added another abstraction because the callback carries no source scene.

mahmoud-elmorabea (approved):

PR #400, opened 2026-08-25
test: add Flutter scene notification E2E (MBL-2303)

Shahroz16:
## Summary

- build and install the scene-configured SwiftPM sample
- tap a real simulator notification from a terminated app with Maestro
- send an HTTPS destination that only reaches Flutter through #399's callback bridge, then assert Settings opens
- extend the existing Xcode 27 nightly/manual lane without a stable-Xcode matrix or duplicate build

The runner hard-gates on notification authorization, waits for Maestro to reach the Home screen before injection, and classifies authorization, injection, tap, and destination failures separately. It uses placeholder workspace configuration and restores local ignored configuration plus the lockfile on exit.

## Validation

- local Xcode 26.6 / iOS 26.5 positive run passed with Maestro 2.8.0: authorization, system notification presentation, tap, and destination-specific Flutter assertion
- negative control passed: with the four #399 bridge source files restored exactly to #399's base commit, the notification was tapped but the Flutter destination assertion failed
- Bash syntax, ShellCheck, YAML/JSON parsing, actionlint, and diff checks passed
- Maestro 2.8.0 archive URL and SHA-256 were verified against the GitHub release asset

This validates simulator presentation, tap handling, and client routing. It does not claim real APNs/FCM token registration or Customer.io backend sent/delivered/opened evidence.


apps/flutter_sample_spm/.maestro/run_scene_push.sh
```diff
@@ -0,0 +1,171 @@
...
+if [[ -z "$device_id" && -n "$simulator_name" ]]; then
+  device_id="$(xcrun simctl list devices available -j | jq -r --arg name "$simulator_name" \
+    '[.devices[][] | select(.name == $name)][0].udid // empty')"
+fi
+if [[ -z "$device_id" && -z "$simulator_name" ]]; then
+  device_id="$(xcrun simctl list devices booted -j | jq -r \
+    '[.devices[][] | select(.state == "Booted") | select(.name | startswith("iPhone"))][0].udid // empty')"
+fi
+if [[ -z "$device_id" ]]; then
+  simulator_name="${simulator_name:-iPhone 17 Pro}"
```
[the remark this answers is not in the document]
Shahroz16: Valid. Fixed in 6fab492 by booting the resolved device after every selection path, before bootstatus. I verified the exact path with a shutdown named iPhone Air simulator; the runner booted it and completed the full system-notification tap and destination assertion.

apps/flutter_sample_spm/.maestro/scene_push_prepare.yaml
```diff
@@ -0,0 +1,24 @@
...
+    visible: "Custom Event Button"
+    timeout: 10000
+- scrollUntilVisible:
+    element: "Show Push Prompt Button"
+    direction: DOWN
+    timeout: 10000
+- tapOn: "Show Push Prompt Button"
+- tapOn:
+    text: "Allow"
+    optional: true
```
mrehan27: The `optional: true` on the permission grant is what let this run report a passing prepare stage while the app was still unauthorized — the tap WARNED with "Element not found: Text matching regex: Allow", the flow continued, and `simctl push` was then rejected with `UNErrorDomain 2003 — Repository could not save notification. Source is not authorized.`

Nothing in this flow asserts that authorization actually ended in `granted`, so an unauthorized device produces a misleading failure 75 seconds later instead of an immediate, accurate one.

The sample already gives us a deterministic in-app signal to assert on rather than relying on Maestro's iOS permission handling: `_showPushPermissionStatus` renders "Push notifications are enabled on this device" when the status is granted (dashboard.dart:202-204), and `_requestPushPermission` shows the same text in a snackbar after a successful grant (dashboard.dart:219-220). Asserting on that string after the prompt interaction fails fast and points at the real cause.

Worth confirming separately whether `launchApp: permissions: notifications: allow` is a no-op on iOS in Maestro 2.6.0 — if it is, the system prompt path is the only one that works and the "Allow" tap needs a real wait rather than `optional: true`.

mrehan27: **P1** — `optional: true` turns a hard precondition into a warning. In run 32874482595 it WARNED with `Element not found: Text matching regex: Allow`, the flow continued unauthorized, and `simctl push` was then rejected with `UNErrorDomain 2003 — Repository could not save notification. Source is not authorized.` — surfacing 75s later as a misleading downstream assertion.

Nothing here asserts authorization actually reached `granted`. The sample already gives a deterministic signal: dashboard.dart:204 and :220 both render `Push notifications are enabled on this device`. Asserting on that fails fast at the real cause.

Also worth confirming whether `launchApp: permissions: notifications: allow` actually applies on iOS in Maestro 2.6.0 — if it's a no-op, the system prompt is the only path and this tap needs a real wait rather than `optional`.

.github/workflows/ios-toolchain-compatibility.yml
```diff
@@ -210,7 +228,31 @@ jobs:
...
+      - name: Record push routing failure
+        if: failure() && steps.push-routing.outcome == 'failure'
+        shell: bash
+        run: |
+          {
+            echo '## Flutter notification routing'
+            echo
+            echo '**Classification:** terminated-notification-routing-failed'
+            echo '**Dependency manager:** SwiftPM'
+            echo '**Next step:** inspect the uploaded Maestro diagnostics; the notification was not tapped or its destination did not reach Flutter'
```
mrehan27: **P1** — wrong for the failure we actually hit. In run 32874482595 the notification was never created, because the app wasn't authorized; this text sends the reader into the routing code instead. Suggest adding the precondition case: `...was never authorized for delivery, was not tapped, or its destination did not reach Flutter`.

apps/flutter_sample_spm/.maestro/run_scene_push.sh
```diff
@@ -0,0 +1,171 @@
...
+    fi
+    sleep 0.5
+  done
+
+  if [[ "$ready" != true ]]; then
+    echo "error: Maestro did not reach the Home screen before notification injection" >&2
+    return 1
+  fi
+
+  xcrun simctl push "$device_id" "$APP_ID" "$payload"
```
mrehan27: **P2** — bare under `set -e`, so a rejection surfaces only as simctl's raw error stream with no indication it came from notification injection. Wrap it with a specific message, and special-case the authorization rejection — that's a harness precondition failure, not a product regression.

apps/flutter_sample_spm/.maestro/scene_push_open.yaml
```diff
@@ -0,0 +1,14 @@
+appId: io.customer.testbed.flutter.spm
+name: Flutter UIScene push routing, open
+---
+- pressKey: Home
+- extendedWaitUntil:
+    visible:
+      id: "NotificationShortLookView"
+    timeout: 30000
```
mrehan27: **P2** — the readiness gate has no margin. In run 32874482595: Home COMPLETED 17:29:56, this 30s wait expired 17:30:27, push result at 17:30:34 — injection landed after the window closed.

The auth failure confounds this (a rejected push plausibly blocks for tens of seconds), so I can't say the happy path races too. But the countdown starts when Maestro executes the step, while injection waits on a `tee`-buffered log line polled at 0.5s — any output lag comes straight out of the window. Either widen it, or have the script confirm the push succeeded before this can time out.

apps/flutter_sample_spm/.maestro/run_scene_push.sh
```diff
@@ -0,0 +1,171 @@
...
+  app_path="$derived_data/Build/Products/Debug-iphonesimulator/Runner.app"
+fi
+test -d "$app_path"
+xcrun simctl uninstall "$device_id" "$APP_ID" >/dev/null 2>&1 || true
+xcrun simctl install "$device_id" "$app_path"
+installed_app=true
+
+maestro --device "$device_id" test .maestro/scene_push_prepare.yaml
+xcrun simctl terminate "$device_id" "$APP_ID"
+run_notification_flow .maestro/scene_push_open.yaml .maestro/fixtures/customerio_scene_settings.apns
```
mrehan27, line 245: **P2** — no negative control, and this workflow already establishes the pattern with `Confirm the AppDelegate-only control reproduces the Xcode 27 launch failure`.

Does this flow fail on the base of #399, without `CustomerIOFlutterDeepLinkRouter`? `DeepLinkUtil.handleDeepLink` falls through callback → `continueNSUserActivity` → `uiKit.open(url:)`, so without the router the SDK still opens `flutter-spm:///settings`, which lands in `scene(_:openURLContexts:)` on `FlutterSceneDelegate`. Whether Flutter's own forwarding then reaches go_router's `/settings` I haven't tested — hence a question, not a claim. But if it passes both with and without #399, the test proves the destination is reachable rather than that this bridge reaches it.

.github/workflows/ios-toolchain-compatibility.yml
```diff
@@ -4,6 +4,7 @@ on:
   pull_request:
     paths:
       - '.github/workflows/ios-toolchain-compatibility.yml'
+      - 'apps/flutter_sample_spm/.maestro/**'
```
mrehan27: **P2** — the filter didn't grow with the coupling surface. This E2E now depends on semantics labels in `apps/flutter_sample_spm/lib/**`, the `flutter-spm` scheme in `ios/Runner/Info-Scene.plist`, and #399's routing in `ios/customer_io/Sources/**` — none of which trigger this workflow on a PR. Renaming a label in dashboard.dart merges green and breaks the nightly the next morning with no attribution. Suggest adding those three paths.

.github/workflows/ios-toolchain-compatibility.yml
```diff
@@ -60,6 +61,14 @@ jobs:
         shell: bash
         run: ./apps/scripts/verify_xcode27_flutter_version.sh --installed
 
+      - name: Install Maestro for scene routing E2E
+        if: matrix.manager == 'SwiftPM'
+        shell: bash
+        run: |
+          export MAESTRO_VERSION=2.6.0
+          curl -Ls https://get.maestro.mobile.dev | bash
```
mrehan27: **P2** — this is the only unpinned third-party code in the file; checkout, setup-ruby, flutter-action and upload-artifact are all SHA-pinned. `MAESTRO_VERSION` pins the artifact but not the installer script fetching it. Either fetch the pinned release tarball with a checksum, or add a comment so the inconsistency reads as deliberate.

(The `mobile-ci-tools/...@main` refs are also unpinned, but those are first-party and presumably intentional.)

.github/workflows/ios-toolchain-compatibility.yml
```diff
@@ -187,8 +196,17 @@ jobs:
...
           test "$(grep -Fc "customerio-flutter-scene-did-become-active token=$run_token" "$callback_log")" -ge 1
 
+      - name: Test terminated notification routing through Flutter
+        id: push-routing
+        if: matrix.manager == 'SwiftPM'
+        shell: bash
+        env:
+          CIO_E2E_APP_PATH: ${{ runner.temp }}/${{ matrix.app }}-scene-xcode-27/Build/Products/Debug-iphonesimulator/Runner.app
+          E2E_DEVICE_ID: ${{ steps.scene-launch.outputs.simulator-udid }}
+        run: apps/flutter_sample_spm/.maestro/run_scene_push.sh
```
mrehan27: **P2** — no `timeout-minutes`. Job budget is 45m and this leg already burned ~23m reaching failure. On a job timeout the step is cancelled rather than failed, so `failure()` is false and `Record push routing failure` is skipped — while `Record launch result` still runs under `always()`. The summary would read `launch-passed` with no routing classification at all. A step-level timeout turns that into a failure the classifier can see.

apps/flutter_sample_spm/.maestro/run_scene_push.sh
```diff
@@ -0,0 +1,171 @@
...
+    build
+
+  app_path="$derived_data/Build/Products/Debug-iphonesimulator/Runner.app"
+fi
+test -d "$app_path"
+xcrun simctl uninstall "$device_id" "$APP_ID" >/dev/null 2>&1 || true
+xcrun simctl install "$device_id" "$app_path"
+installed_app=true
+
+maestro --device "$device_id" test .maestro/scene_push_prepare.yaml
```
mrehan27: **P2** — prepare runs without `--debug-output`, unlike `run_notification_flow`, yet the failure summary tells the reader to inspect Maestro diagnostics. That gap bit this run specifically: the WARNED permission tap — the actual root cause — was in prepare, and the only debug directory produced was `flutter_sample_spm-maestro-scene_push_open`.

mrehan27 (commented):
Good approach — Maestro + `simctl push` against the scene host is the right shape, and reusing the nightly lane instead of adding a matrix is the right call. Cleanup, config restore, and the classification-step convention all match the file.

Blocking on one thing: the lane this extends is red on this head and it isn't a flake. [Run 32874482595](https://github.com/customerio/customerio-flutter/actions/runs/32874482595/job/97889034900) failed at the new step because notification authorization was never granted, so `simctl push` was rejected outright and no banner ever existed. The `optional: true` permission tap turned that precondition into a warning. Details inline.

mrehan27 (approved):
Approving the approach — Maestro + `simctl push` against the scene-configured host is the right shape, and extending the existing nightly lane instead of adding a matrix is the right call.

Before merging:

1. **Green SwiftPM lane.** It currently fails on the notification-authorization precondition rather than on routing — see the two P1s. `simctl push` is being rejected with `Source is not authorized`, so no banner is ever created.
2. **Settle the negative-control question** (`run_scene_push.sh:171`). If the flow also passes without #399's router, the test needs rethinking rather than patching — worth knowing before this lands in a daily lane.
3. **Retarget to main after #399 merges**, per the description.

The remaining P2s are worth folding in while you're in here, but they don't block.

apps/flutter_sample_spm/.maestro/scene_push_open.yaml
Shahroz16: Valid. Fixed in 91b07ef. The presentation wait is now 60 seconds, and injection failures are classified at the source. The corrected happy-path run passed locally.

apps/flutter_sample_spm/.maestro/run_scene_push.sh
Shahroz16: Confirmed with a real negative control and fixed in 91b07ef. The custom scheme could reach Flutter without proving #399, so the fixture now uses https://www.flutter-spm.com/settings. The flow passed with #399. With the four bridge source files restored exactly to #399 base commit 2fce8244, the notification was tapped but the final Flutter destination assertion failed as expected.

.github/workflows/ios-toolchain-compatibility.yml
Shahroz16: Valid. Fixed in 91b07ef. The failure summary now names authorization/injection failure separately from tap and Flutter destination failure.

.github/workflows/ios-toolchain-compatibility.yml
Shahroz16: Valid. Fixed in 91b07ef. The PR trigger now includes the sample Dart UI, ios/Runner/Info-Scene.plist, and Customer.io iOS plugin sources in addition to the harness and workflow.

apps/flutter_sample_spm/.maestro/run_scene_push.sh
Shahroz16: Valid. Fixed in 91b07ef. simctl push output is now captured and classified; authorization rejection produces a specific harness-precondition error, while other injection failures get a separate message.

.github/workflows/ios-toolchain-compatibility.yml
Shahroz16: Valid. Fixed in 91b07ef. CI now downloads the pinned Maestro 2.6.0 release archive directly and verifies GitHub's published SHA-256 before extraction; the mutable installer script is no longer used.

apps/flutter_sample_spm/.maestro/run_scene_push.sh
Shahroz16: Valid. Fixed in 91b07ef. The prepare flow now writes its own flattened Maestro debug output under RUNNER_TEMP/flutter_sample_spm-maestro-scene_push_prepare, so permission/precondition failures are included in uploaded diagnostics.

apps/flutter_sample_spm/.maestro/scene_push_prepare.yaml
Shahroz16: Valid. Fixed in 91b07ef. The Allow tap remains optional only for already-authorized reruns, but the flow now has a hard gate on the app exact Push notifications are enabled on this device signal before injection. The corrected local flow passed.

.github/workflows/ios-toolchain-compatibility.yml
Shahroz16: Valid. Fixed in 91b07ef. The routing step now has a 10-minute timeout so cancellation is classified as a step failure instead of silently leaving only the launch result.

apps/flutter_sample_spm/.maestro/scene_push_prepare.yaml
Shahroz16: Follow-up: I also moved the hosted lane from Maestro 2.6.0 to the locally tested 2.8.0 release. The previous 2.6.0 run is evidence that its permission setup was not reliable here; the new exact app-status gate still fails fast if authorization is not actually granted.

.github/workflows/ios-toolchain-compatibility.yml
Shahroz16: Follow-up: the pinned archive is now Maestro 2.8.0, matching the successful local positive and negative-control runs. Its GitHub release SHA-256 was independently downloaded and verified before push.

.github/workflows/ios-toolchain-compatibility.yml
```diff
@@ -4,6 +4,7 @@ on:
   pull_request:
     paths:
       - '.github/workflows/ios-toolchain-compatibility.yml'
+      - 'apps/flutter_sample_spm/.maestro/**'
   schedule:
     # Native runs one hour earlier; Expo runs one hour later.
```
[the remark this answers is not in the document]
Shahroz16: Intentional. This Xcode 27 lane is expensive, so SDK and sample changes are covered by the nightly schedule and manual dispatch rather than adding per-PR Xcode 27 runs to every change. The pull-request filter is deliberately limited to changes to this workflow or its Maestro harness, where immediate validation is required. Regular CI still builds the samples on stable Xcode.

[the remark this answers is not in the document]
Shahroz16: Follow-up after a fresh adversarial review: the broad SDK/sample path expansion would still be contrary to the nightly-only cost decision, but the AppDelegate now contains test-harness permission wiring. I added that exact path in ec62f99 so changes to the hook run this PR check immediately; other SDK/source changes remain covered nightly or through manual dispatch.

PR #398, opened 2026-08-21
fix: restore Flutter package publishing

Shahroz16:
## Summary

- Move the customer-facing CocoaPods guide to Dart’s conventional `doc/` directory.
- Keep the internal Xcode CI note under `.github/docs/`, outside the published package.
- Run `flutter pub publish --dry-run` directly in PR CI so package validation cannot be skipped when the local and published versions match.
- Add Flutter 3.47 standard platform exclusions so `flutter pub get` does not migrate a tracked file before validation.

## Incident and recovery

The 4.3.0 GitHub release and tag were created successfully, but pub.dev rejected the package dry run because the new top-level `docs/` directory produced a layout warning. Pub.dev therefore remains on 4.2.1.

The previous PR check did not catch this because the publisher action saw local 4.2.1 equal pub.dev 4.2.1 and skipped its dry run. This PR removes that false-positive path. Its `fix:` commit should produce and publish 4.3.1 after merge without rewriting the immutable 4.3.0 tag.

## Validation

- `flutter pub publish --dry-run`: 0 warnings
- Hosted Flutter 3.47.1 `test_publish`: passed
- `actionlint .github/workflows/test.yml`: passed
- Documentation contents are unchanged; only their locations changed.

---

> [!NOTE]
> **Medium Risk**
> Changes package layout and CI publishing validation, but does not modify runtime SDK code or the actual release publisher workflow.
> 
> **Overview**
> Restores Flutter package publishing after the 4.3.0 release was rejected by pub.dev.
> 
> Moves the customer-facing CocoaPods guide from `docs/` to Dart's conventional `doc/` directory and updates the README link, resolving the pub.dev layout warning that blocked 4.3.0. The internal Xcode CI note stays under `.github/docs/` so it is excluded from the published package.
> 
> Replaces the `k-paxian/dart-package-publisher` action in PR CI with a direct `flutter pub publish --dry-run` command. The previous action skipped validation when the local version matched the published version, which is why the `docs/` warning was not caught before merge.
> 
> Adds Flutter 3.47 standard platform exclusions to `analysis_options.yaml` so `flutter pub get` does not modify tracked files before validation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 260a39ae5a77c578a6537328b2f5a8526fa55702. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #403, opened 2026-08-31
chore: resolve the published SDK version from the release job and validate before tagging

mahmoud-elmorabea:
Fixes MBL-2324

## Problem

Two independent defects in the Flutter release pipeline.

**1. Post-release sample-app builds use the *previous* release version.**

`reusable-build-sample-apps.yml` resolved the SDK version with `git describe --tags --abbrev=0`. That step runs on the commit that *triggered* the release, but semantic-release tags a **new child commit** (`chore: prepare for X [skip ci]`). `git describe` only walks ancestors, so it deterministically returns the previous release. Verified in this repo:

| tag | tagged commit | `git describe` from its parent |
| --- | --- | --- |
| 4.3.0 | 137a72c | 4.2.1 |
| 4.3.1 | 2fce824 | 4.3.0 |
| 4.4.0 | 80d89f5 | 4.3.1 |

`apps/scripts/update_sample_app_sdk_version.dart` then writes that stale value into the sample app's `pubspec.yaml` as an exact pin. One cause, three symptoms:

1. iOS Fastlane SPM conflict — `'customer_io-4.3.1' depends on 'customerio-ios' 4.7.5 and root depends on 'customerio-ios' 4.7.6`
2. Android Dart compile failure — `Member not found: 'liveActivities'` (the sample app uses APIs only in the version being released)
3. `flutter pub get` failure when the stale tag was never published (4.3.0 is tagged but is not on pub.dev)

**2. The git tag is cut before publish validation runs.**

`k-paxian/dart-package-publisher` fails on *any* `pub publish` dry-run warning, and it only runs in `deploy-to-pub-dev`, i.e. after `deploy-git-tag` has already created the tag. When #394 added a top-level `docs/` directory, the dry run emitted `Rename the top-level "docs" directory to "doc"` and the publish step aborted with `Dry 🏃 Failed, skip real publishing` — but tag `4.3.0` already existed, and that orphan tag then broke the following release run. The content issue was fixed by #398; the ordering was not.

## What changed

**Version resolution**

`LATEST_TAG` had two jobs conflated into one value, so this splits them rather than redefining it:

- `LATEST_TAG` keeps its current meaning — the last tag reachable from `HEAD`, always from `git describe`. Its one ancestry consumer, `COMMITS_AHEAD_COUNT` (`git rev-list $LAST_TAG..HEAD --count`), is a genuine ancestry question and is untouched on every code path, release included.
- New `SDK_RELEASE_VERSION` is the *published* version the sample app is pinned to: `${SDK_VERSION_INPUT:-$LATEST_TAG}`. It feeds the three consumers that all sit behind `use_latest_sdk_version` — the Fastlane `version_name`, the `.env` `SDK_VERSION` the Dart pin script reads, and the Slack `APP_SDK_BUILD_VERSION`.

Because the fallback triggers on an empty input rather than on event name, PR, push and `workflow_dispatch` builds resolve exactly as they do today; only the release path supplies a value.

- `reusable-build-sample-apps.yml`: new optional `sdk_version` input; both `Capture Git Context` steps (Android and iOS) updated. The two call sites turned out to be symmetric — identical `LATEST_TAG` consumer sets on both.
- `build-sample-app-for-sdk-release.yml`: threads `sdk_version` through from both `workflow_call` and `workflow_dispatch`.
- `tag-and-deploy.yml`: `publish-sample-apps-public-builds` now also `needs` `deploy-git-tag` and passes `sdk_version: ${{ needs.deploy-git-tag.outputs.new_release_version }}`. That output already existed; `needs` only exposes outputs of directly-listed jobs, so growing the `needs` list is the whole change. `deploy-to-pub-dev` already depended on `deploy-git-tag`, so job ordering is unchanged.

The version reaches the build through a nested reusable-workflow `uses:` chain — `tag-and-deploy.yml` → `build-sample-app-for-sdk-release.yml` → `reusable-build-sample-apps.yml` — so the input has to be declared and forwarded at each hop.

**Validate before tagging**

- `tag-and-deploy.yml`: `deploy-git-tag` now runs `flutter pub publish --dry-run` *before* the semantic-release step. A validation failure now fails the job before any tag exists.

Two notes on that choice:

- It is a plain `flutter pub publish --dry-run` rather than `dart-package-publisher` with `dryRunOnly: true`. The action short-circuits on `LOCAL_PACKAGE_VERSION == REMOTE_PACKAGE_VERSION` *before* it reaches the dry run — and on `main` before a release, `pubspec.yaml` still holds the last published version (4.4.0, matching pub.dev today), so the action would skip the dry run entirely and pass vacuously. The bare dry run reproduces the same gate faithfully: reintroducing a top-level `docs/` directory locally makes it exit 65 with the exact warning from the incident.
- The dry run is a step in the existing job rather than a new upstream job, so the job's existing `if: failure()` Slack notification still fires. Its message was updated to `*validate package & create git tag*`.

## Verification

- Both premises reproduced locally: the tag-on-child-commit table above, and `flutter pub publish --dry-run` exiting 65 on the `docs/` warning (0 warnings on `main` as-is).
- All three run shapes simulated against this checkout: release (input set) keeps `LATEST_TAG=4.4.0` with `COMMITS_AHEAD` unchanged while `SDK_RELEASE_VERSION` takes the passed version; PR/push (input empty) resolves both to `4.4.0`, identical to today; a tagless repo survives `bash -eo pipefail`, which is why the `git describe` call carries `|| true`.
- Bash lazy expansion of `${VAR:-...}` confirmed: `git describe` still runs once for `LATEST_TAG`, and the input short-circuits only the `SDK_RELEASE_VERSION` default.
- All workflow YAML re-parses cleanly.
- The release path itself can only be confirmed on the next real release — a PR cannot exercise a `push: [main]` workflow's `uses:` chain.

## Out of scope

The existing orphan tag `4.3.0` is deliberately left in place.

### DIFF
--- a/.github/workflows/build-sample-app-for-sdk-release.yml
+++ b/.github/workflows/build-sample-app-for-sdk-release.yml
@@ -2,7 +2,19 @@ name: Publish test apps for SDK release
 
 on:
   workflow_dispatch:
+    inputs:
+      sdk_version:
+        description: "Published SDK version to build sample apps with. Falls back to the latest git tag when empty."
+        required: false
+        type: string
+        default: ""
   workflow_call:
+    inputs:
+      sdk_version:
+        description: "Published SDK version to build sample apps with. Falls back to the latest git tag when empty."
+        required: false
+        type: string
+        default: ""
 
 jobs:
   build-primary-app:
@@ -12,4 +24,5 @@ jobs:
       sample_app_name: flutter_sample_spm
       cio_workspace_name: "Mobile: Flutter"
       is_primary_app: true
+      sdk_version: ${{ inputs.sdk_version }}
     secrets: inherit
--- a/.github/workflows/reusable-build-sample-apps.yml
+++ b/.github/workflows/reusable-build-sample-apps.yml
@@ -37,6 +37,11 @@ on:
         type: boolean
         required: false
         default: false
+      sdk_version:
+        description: "Published SDK version to build sample apps with. Falls back to the latest git tag when empty."
+        type: string
+        required: false
+        default: ""
     outputs:
       APP_VERSION_NAME:
         description: "The version name of the built app"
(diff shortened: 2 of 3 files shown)


Shahroz16 (approved):
Core release fix looks good, including the pre-tag dry run. One hardening follow-up: the manual `sdk_version` input is interpolated directly into shell commands in the reusable workflow. Since those jobs inherit secrets, it would be safer to validate the input as a release version and reference `$SDK_RELEASE_VERSION` from the shell instead of embedding `${{ env.SDK_RELEASE_VERSION }}`. Manual dispatch is limited to trusted repo access, so I do not think this needs to block the PR.

PR #408, opened 2026-09-11
feat(inbox): let apps configure visual inbox accessibility labels

mahmoud-elmorabea:
Exposes the visual notification inbox accessibility labels through the Flutter config, so host apps supply their own localized strings instead of the SDK shipping English text.

Fixes MBL-2366

## Notes
- iOS needs no native change: the `inApp` map already reaches `MessagingInAppConfigBuilder.build(from:)`, which parses these keys.
- Android builds the labels and converts the `{count}` template into the `(Int) -> String` closure the SDK takes — the channel carries data, not functions.
- `toMap()` omits unset labels rather than sending explicit nulls, and `InAppConfig` omits the labels map entirely when it serializes empty, so "configured nothing" and "configured an empty object" stay distinguishable natively.
- A `bellWithUnreadCount` missing its `{count}` placeholder is announced verbatim with no count. Dart warns about it: Android would log at debug, which the default error level discards, and iOS does not check at all, so neither native layer can report it where the developer would see it.
- Pins bumped to the native releases that added this config: iOS 4.8.0, Android 4.21.1 — `pubspec.yaml`, `android/build.gradle`, `ios/customer_io/Package.swift`, and both sample apps (the SPM sample's pbxproj and the cocoapods sample's Podfile). Both podspecs read `native_sdk_version` from pubspec, so they follow automatically.
- The sample app supplies labels at its single `initialize()` call. No other sample changes: the inbox screens already exist on main, and the empty state is reached by logging in with a profile that has no inbox messages.

## Verification
- `flutter test` — 122 passed (11 new)
- `flutter analyze` — no issues in `lib/` or `test/`
- `./scripts/check_api_changes.sh` — no API changes detected
- `./gradlew :customer_io:testDebugUnitTest` against Android SDK 4.21.1 — 7 tests, 0 failures (confirmed in the JUnit XML, since the plugin's Test tasks do not fail the build on their own)

### DIFF
--- a/lib/config/notification_inbox_accessibility_labels.dart
+++ b/lib/config/notification_inbox_accessibility_labels.dart
@@ -1,0 +1,70 @@
+/// Host-provided accessibility labels for the Visual Notification Inbox UI.
+///
+/// The SDK ships no text of its own in the visual inbox — the empty state is an icon and the
+/// loading state is a spinner — so accessibility labels are the one place a string is still
+/// needed. Because the SDK cannot know your app's language, every label is optional and null by
+/// default, and an omitted label leaves that element unlabeled rather than falling back to
+/// English. Pass strings already localized for the user's language.
+///
+/// ```dart
+/// CustomerIO.initialize(
+///   config: CustomerIOConfig(
+///     cdpApiKey: '...',
+///     inAppConfig: InAppConfig(
+///       siteId: '...',
+///       accessibilityLabels: NotificationInboxAccessibilityLabels(
+///         bell: context.l10n.inboxBell,
+///         bellWithUnreadCount: context.l10n.inboxUnread, // "{count} olasta aviseringar"
+///         loadingIndicator: context.l10n.inboxLoading,
+///         emptyState: context.l10n.inboxEmpty,
+///       ),
+///     ),
+///   ),
+/// );
+/// ```
+class NotificationInboxAccessibilityLabels {
+  /// Label for the inbox bell button. Also used when the bell shows an unread badge but
+  /// [bellWithUnreadCount] is not provided. null → the bell is announced as an unnamed button
+  /// (it stays focusable and tappable; hiding it would leave screen reader users no way in).
+  final String? bell;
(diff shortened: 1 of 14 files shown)


lib/config/notification_inbox_accessibility_labels.dart
```diff
@@ -0,0 +1,96 @@
...
+  /// `{{count}}`, `%d`) matches nothing, so the template is announced verbatim — braces and all —
+  /// and the count is never spoken. The check lives in Dart rather than in either native layer
+  /// because neither can report it usefully: Android logs at debug, which the default error level
+  /// discards, and iOS does not check at all. Warning here reaches the developer's console on both
+  /// platforms, whatever the SDK log level. It warns rather than throws: a label typo degrades an
+  /// announcement, it must not fail initialization.
+  void _warnIfCountPlaceholderMissing() {
+    final String? template = bellWithUnreadCount;
+    if (template != null && !template.contains(countPlaceholder)) {
+      debugPrint(
```
Shahroz16, line 94: Could we gate this warning with `kDebugMode` as a small follow-up? `debugPrint` also writes in release builds, while the other Dart diagnostics and the React Native equivalent are limited to development builds.

Shahroz16 (approved):
Checked the native config mapping, focused tests, and Android build. The Xcode 27 SwiftPM notification-permission failure is also present on main: https://github.com/customerio/customerio-flutter/actions/runs/34937959825.

lib/config/notification_inbox_accessibility_labels.dart
mahmoud-elmorabea: Good catch — done in 14682f7, in this PR rather than a follow-up.

You were right about the precedent: every `debugPrint` in `lib/` is already wrapped in `kDebugMode`, including the location and geofence ones. This was the only unguarded one. Gating it also lines Flutter up with the React Native equivalent, which is `__DEV__`-gated.

Leaving your approval as is.

PR #410, opened 2026-09-17
feat(in-app): allow overriding the in-app message color scheme

mahmoud-elmorabea:
Exposes the native SDKs' in-app color scheme override so a host app with its own appearance setting can pin the light or dark variant instead of always following the device. Mirrors the React Native change in customerio-reactnative#660.

- `InAppConfig.colorScheme`, and `CustomerIO.inAppMessaging.setColorScheme` for changing it at runtime, both taking the new `CioColorScheme` enum.
- Named `CioColorScheme` because Material already exports `ColorScheme` and hosts import both — unprefixed it would be an ambiguous import in most apps.
- iOS needed no change on the init path: `MessagingInAppConfigBuilder.build(from:)` already parses the key, so only the Android half of that path was missing. The runtime setter is new on both.
- Messages already on screen re-theme in place; that is native behavior this only reaches.

Contributes to MBL-2479


mrehan27 (approved):
Color scheme wiring looks right — same wire values as customerio-reactnative#660, and the strict setter / lenient init split is consistent on both platforms.

Both native pins are current: iOS 4.8.1, Android 4.21.1.

pubspec.yaml
```diff
@@ -43,5 +43,5 @@ flutter:
         pluginClass: CustomerIOPlugin
       ios:
         pluginClass: CustomerIOPlugin
-        native_sdk_version: 4.8.0
+        native_sdk_version: 4.8.1
```
Shahroz16: Could we explain why this feature PR also moves every iOS pin from 4.8.0 to 4.8.1, or split that dependency update out? The color-scheme APIs are already present in 4.8.0, while 4.8.1 contains unrelated push and dependency changes. Keeping that behavior change in the wrapper feature makes regressions harder to attribute and leaves React Native on 4.8.0.

mahmoud-elmorabea: Any reason to split this into two releases?

Shahroz16: Nope agent went rogue, didn't know what to test because there were multiple releases involved 😅

Shahroz16 (approved):

